PR #6, opened 2012-02-29
Change to invoke gradle via Java rather than gradle/gradle.bat

staale:
Using the gradle plugin together with the gerrit-trigger plugin, and with a remote windows jenkins build server revealed som problems related to windows command line parsing. The gerrit trigger adds several build parameters to a build, including the name and email off the person submitting a patch.

The parameter GERRIT_PATCHSET_UPLOADER gets the value "Firstname Lastname <user@example.com>". When this gets added as a command line argument to the gradle.bat file, windows interpets the "<" as a pipe from file, and fails to run the bat command since the file user@example.com is not found. In windows, command line arguments can be escaped with "^", however, that only worked for invoking the gradle bat script, and then caused different problems within the bat script. Even double escaping with "^^^<" didn't seem to fix the problem.

What this change does, is to look in the build environment for "JAVA_HOME", which is set by selecting a JDK for the build. It assumes that gradle is started by adding "gradle-launcher*.jar" to the class path from the lib folder off gradle home, and invoking "org.gradle.launcher.GradleMain". I have tested this for gradle-1.0-milestone-7, and the same will work for milestone-8. This change completly ignores the gradle wrapper setting, which I am not even sure what does. Maybe it's related to an old version off gradle.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ work
 *.iml
 *.ipr
 *.iws
+.idea
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
 
     <groupId>com.zenika.jenkins-ci.plugins</groupId>
     <artifactId>gradle</artifactId>
-    <version>1.12-SNAPSHOT</version>
+    <version>1.12</version>
     <packaging>hpi</packaging>
     <name>Jenkins Gradle plugin</name>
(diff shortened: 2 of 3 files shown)


gezerk:
Rejecting since the PR is very old and the Gradle ecosystem has changed a lot. Planning to rework the plugin to pull in some of the PR's around the wrapper.  We can fix any problems with Windows/Gerrit interaction then.

PR #12, opened 2012-09-05
Set GRADLE_USER_HOME all the time

quidryan:
We have our home directory to read-only, so it detects things where builds are writing to a common location. I found that gradle will write to ~/.gradle by default, which is really bad for a multi-executor slave in Jenkins. Just setting the environment variable localized files to the workspace.

### DIFF
--- a/src/main/java/hudson/plugins/gradle/Gradle.java
+++ b/src/main/java/hudson/plugins/gradle/Gradle.java
@@ -183,6 +183,9 @@ public class Gradle extends Builder implements DryRun {
             env.put("GRADLE_HOME", ai.getHome());
         }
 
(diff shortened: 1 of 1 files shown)


daniel-beck:
@quidryan I know this issue is old, but:

> "Gradle's advanced cache (which is what we're talking about sharing) uses locks. By sharing the same cache between executors, you're going to be locking parts of the cache which will block other executors. This is an unnecessary slow down"

Just define a node-level variable `GRADLE_USER_HOME` with a value that includes the variable `$EXECUTOR_NUMBER`. This will result in executor-specific Gradle caches (therefore no lock contention) with only minor reduction in cache effectiveness. Problem solved.

I have serious doubts anyone really needs to set per-project Gradle caches.

PR #14, opened 2012-12-14
Fix gerrit parameter handling for Windows

gluck:
Gerrit plugin adds many build variables, first one being:
-DGERRIT_PATCHSET_UPLOADER=\"Foo Bar\" foo.bar@acme.com
(with the back-slashes, as required on unix)

This causes issues with gradle depending on how it's escaped, either:
- gradle.bat fails if the first argument isn't 'quote-balanced' (and batch files don't recognize quote escaping through '\')
- gradle fails in jdk6 but not in jdk7 because of a java command line parsing change:
  http://gluck-md.blogspot.fr/2012/12/jdk7-surprise.html

Patch does 2 things:
- remove custom escaping and use the common jenkins one (used by Ant)
  (that alone isn't enough as it doesn't work for jdk below 7)
- add a configuration to bypass the startup script (both gradle.bat & gradlew.bat) and call directly java

Basically does what this pull request was meant to, but keeping gradlew support:
https://github.com/jenkinsci/gradle-plugin/pull/6

### DIFF
--- a/src/main/java/hudson/plugins/gradle/Gradle.java
+++ b/src/main/java/hudson/plugins/gradle/Gradle.java
@@ -11,8 +11,11 @@ import org.jenkinsci.lib.dryrun.DryRun;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.StaplerRequest;
 
+import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


wolfs:
@gluck Thank you for your PR and the work you put into it. I am sorry that this PR has been unanswered for so long.
On current `master` I already integrated using the default Jenkins escaping. Moreover, it is now possible to disable passing all parameters as system properties to Gradle. I will close this PR. If you still think there is some work left to do feel free to re-open it again.

PR #22, opened 2014-05-06
Defined 2 custom view job filters 'All Gradle Jobs' and 'Gradle Jobs'.

fritaly:
The first filter will alter the selection of a list view to include all
the jobs using the gradle builder. The second filter will alter the
selection of a list view to only keep the jobs using the gradle builder.

### DIFF
--- a/src/main/java/hudson/plugins/gradle/AllGradleJobsFilter.java
+++ b/src/main/java/hudson/plugins/gradle/AllGradleJobsFilter.java
@@ -1,0 +1,55 @@
+package hudson.plugins.gradle;
+
+import hudson.Extension;
+import hudson.model.TopLevelItem;
+import hudson.model.Descriptor;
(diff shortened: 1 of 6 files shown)


wolfs:
@fritaly Thank you for your PR and the work you put into it! I am very sorry for the late response.
Are you still interested in including this work in the Gradle plugin?

fritaly:
Hi @wolfs, I had completely forgotten about this PR :) I have since moved to another company and don't use Jenkins any more so I personally don't need this PR to be merged. 

I'll let you decide if it can be useful to others. At that time, I implemented those view filters to easily monitor Gradle builds in Jenkins: it helped me to detect jobs that I hadn't created personally on the server.

wolfs:
Closing this now since I won't pick it up. If anybody else feels like this is a useful feature, please re-open.

PR #24, opened 2015-07-21
[JENKINS-25634] Add ability to ignore error code from Gradle build.

scmbuildguy:
This commit adds a checkbox to the plugin UI, which when checked, the build step will ignore the output of the Gradle build.

Successfully tested by checking and unchecking new checkbox while executing Gradle build steps which always fail.

### DIFF
--- a/src/main/java/hudson/plugins/gradle/Gradle.java
+++ b/src/main/java/hudson/plugins/gradle/Gradle.java
@@ -30,10 +30,11 @@ public class Gradle extends Builder implements DryRun {
     private final boolean makeExecutable;
     private final boolean fromRootBuildScriptDir;
     private final boolean useWorkspaceAsHome;
+    private final boolean ignoreBuildResult;
 
     @DataBoundConstructor
     public Gradle(String description, String switches, String tasks, String rootBuildScriptDir, String buildFile,
-                  String gradleName, boolean useWrapper, boolean makeExecutable, boolean fromRootBuildScriptDir, boolean useWorkspaceAsHome) {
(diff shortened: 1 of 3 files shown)


rbywater:
Out of interest, what's the use case for ignoring the output of Gradle?

pivotalbuildguy:
Use case is when running Findbugs through Gradle plugin, if there are any issues generated then the Gradle build fails and which then stops all other future build steps from happening. This allows a user to run build steps after a 'failed' Gradle build.

rbywater:
I'm not against this change per se, but presumably if that is the behaviour you are after then you could use the "ignoreFailures" property on the FindBugs task to perform the same action?

pivotalbuildguy:
Ok, you got me there :) Yes, 'ignoreFailures' works for FindBugs. It was my testcase as I am doing Java and the original request came from a Android Lint user. This feature is mostly for convenience of not having to add too many command line switches to a build if there is a failure.

We could take this discussion a step further and say this feature is not needed, then users must use the ignoreFailures property, however, I do not know if all gradle plugins have the functionality. I know 'java', 'findbugs', and others have it; if a user created a custom plugin, they would have to be able to have the functionality as well.

wolfs:
Thank you very much for your pull request and sorry about the late response.

Like @rbywater already mentioned, this should better be handled at the Gradle build file level and not at the Jenkins level. You can easily check for one of the [environment variables](https://wiki.jenkins-ci.org/display/JENKINS/Building+a+software+project#Buildingasoftwareproject-JenkinsSetEnvironmentVariables) set by Jenkins and determine in your build script if you are running on Jenkins and set the flags accordingly.

PR #26, opened 2015-08-23
JENKINS-26287 Prevent the console-outline to overlap the console output

youjenli:
JENKINS-26287 Prevent the console-outline block to overlap the console output in case of long task paths.

Add stylesheet and unorder list element for console-outline table  to
prevent it overlapping the console output in case of long task paths.


rompic:
would love to see this merged.

gezerk:
Looks good to me.

PR #25, opened 2015-08-22
Add new build step option "transferEnvironmentVariables".

youjenli:
Provide a build step checkbox option for users to decide whether Gradle
plugin should transfer sensative environment variables to Gradle command
line as jvm system variables or not.
In order to comply backward compatibility, default value of this option
is true.

PR #27, opened 2016-02-17
[JENKINS-17523] feat: Enable passing job parameter as properties

jhinrichsen:
As of version 1.24 of the Gradle plugin, all job parameters are passed as JVM system properties via '-D'. This change keeps the default behaviour, but allows to pass job parameters via regular Gradle properties, i.e. '-P' so that they are immediately available in the Gradle DSL.

To pick one or the other, the job config GUI contains an additional checkbox with a corresponding help text (as of now, in english only).

Fixes JENKINS-17523, JENKINS-17523

### DIFF
--- a/src/main/java/hudson/plugins/gradle/Gradle.java
+++ b/src/main/java/hudson/plugins/gradle/Gradle.java
@@ -30,10 +30,12 @@ public class Gradle extends Builder implements DryRun {
     private final boolean makeExecutable;
     private final boolean fromRootBuildScriptDir;
     private final boolean useWorkspaceAsHome;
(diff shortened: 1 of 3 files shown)


jhinrichsen:
Agree. There's some Jira issues and PR w/ regard to finer control over what gets passed at all, but that is too far away from my setup. I only bind credentials that are needed, so there's no uncontrolled flooding of unwanted environt variables, and no need to filter stuff (Add new build step option "transferEnvironmentVariables". #25)

Good to see the plugin is picking up speed!

gezerk:
Thanks for the PR!

PR #35, opened 2016-08-13
[FIXED JENKINS-37394] Add @Symbol annotations for step and tool.

abayer:
Specifically added them to the Descriptors.

cc @reviewbybees

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,6 +37,8 @@ sourceCompatibility = '1.6'
 
 dependencies {
   compile 'org.jenkins-ci.lib:dry-run-lib:0.1'
+  compile 'org.jenkins-ci:symbol-annotation:1.3'
+  jenkinsPlugins 'org.jenkins-ci.plugins:structs:1.3@jar'
 
   testCompile 'org.spockframework:spock-core:0.7-groovy-1.8'
   jenkinsTest 'org.jenkins-ci.main:jenkins-test-harness:2.8@jar'
--- a/build.gradle
+++ b/build.gradle
@@ -37,9 +37,8 @@ sourceCompatibility = '1.6'
 
 dependencies {
   compile 'org.jenkins-ci.lib:dry-run-lib:0.1'
-  jenkinsPlugins('org.jenkins-ci.plugins:structs:1.3@jar') {
-    transitive = true
-  }
+  compileOnly 'org.jenkins-ci:symbol-annotation:1.3'
+  jenkinsPlugins 'org.jenkins-ci.plugins:structs:1.3@jar'
 
   testCompile 'org.spockframework:spock-core:0.7-groovy-1.8'
   jenkinsTest 'org.jenkins-ci.main:jenkins-test-harness:2.8@jar'
--- a/build.gradle
+++ b/build.gradle
@@ -37,8 +37,9 @@ sourceCompatibility = '1.6'
 
 dependencies {
   compile 'org.jenkins-ci.lib:dry-run-lib:0.1'
-  compile 'org.jenkins-ci:symbol-annotation:1.3'
-  jenkinsPlugins 'org.jenkins-ci.plugins:structs:1.3@jar'
+  jenkinsPlugins('org.jenkins-ci.plugins:structs:1.3@jar') {
+    transitive = true
+  }
 
   testCompile 'org.spockframework:spock-core:0.7-groovy-1.8'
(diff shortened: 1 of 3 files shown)


wolfs:
It would be nice to have an integration test for that. Is this easily possible with the current JenkinsRule infrastructure?

abayer:
Once https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/16
lands, we can do a `tool` test. Or we can use `SymbolLookup` now...

build.gradle
```diff
@@ -37,6 +37,8 @@ sourceCompatibility = '1.6'
 
 dependencies {
   compile 'org.jenkins-ci.lib:dry-run-lib:0.1'
+  compile 'org.jenkins-ci:symbol-annotation:1.3'
+  jenkinsPlugins 'org.jenkins-ci.plugins:structs:1.3@jar'
```
jglick: Uh, should suffice to add the `structs` dependency without specifying `symbol-annotation`, no?

abayer: Surprisingly, no! Gradle, man, I dunno.

jglick: Hmm. Just check the output `*.hpi` file carefully to make sure it

* declared a plugin dependency on `structs`
* does _not_ include `WEB-INF/lib/symbol-annotation.jar`

jglick:
:bee: assuming the artifact is correct despite the odd behavior of `gradle-jpi-plugin` or whatever it is called.

build.gradle
abayer: All I know is I had compilation errors without it. *shrug*

wolfs: Try using
```groovy
jenkinsPlugins 'org.jenkins-ci.plugins:structs:1.3@jar' {
    transitive = true
}
```

When using `@jar` `transitive` is automatically set to false. You will now pull in transitive `hpi` files, if the plugins depends on other plugins. I think something to do this automatically should be incorporated in the `gradle-jpi-plugin`.

abayer: Hrm - changed that and it compiled, but then tests failed. Oy!

abayer: Fixed!

abayer:
Ok, what finally got both compilation and tests to work was a `compileOnly` dependency on `symbol-annotation`, a la Maven `provided` scope. Sigh.

wolfs:
At some point someone should revisit the `gradle-jpi-plugin` to fix this classpath mess. Could you give me a pointer to how this happens in Maven land?

abayer:
I think in Maven, the dependency is on the hpi, not the jar, so it's
basically a war dependency, picking up other jar dependencies of the hpi
from inside it...

jglick:
No, for Maven the dependency is on the primary artifact, of type `jar`, but the plugin does some analysis of the dependency trail to decide which plugin dependencies to add to the manifest and which JARs to bundle.

jglick:
re:bee:

wolfs:
What is the Cloudbees process for this PR? Should I merge it or are you going to?

abayer:
@reviewbybees done

abayer:
@wolfs Up to you whether to merge it. =)

wolfs:
@abayer With this change I still can't use Gradle in Pipeline since it is no `SimpleBuildStep`, right?

abayer:
@wolfs Correct. And more importantly, the gradle step isn't a durable task step, so it shouldn't be used in Pipeline even if it is changed to be a `SimpleBuildStep`.

wolfs:
@abayer That sound interesting. Would it be difficult to add a durable task step for the Gradle plugin? Is there a good example how to do that?
Regarding the features of durable task - can this also be used without using pipeline? E.g. I have Gradle running on a Slave, restart Jenkins, and then the Job reattaches and shows Gradle still running?

abayer:
@wolfs I'll defer to @jglick to explain what's possible in detail, but no, durable tasks only work that way within Pipeline jobs.

jglick:
> Would it be difficult to add a durable task step for the Gradle plugin? Is there a good example how to do that?

Discussion better left to [https://issues.jenkins-ci.org/browse/JENKINS-27393](https://issues.jenkins-ci.org/browse/JENKINS-27393).

wolfs:
@abayer I merged this but I do not know how to invoke these custom symbols from the Pipeline script. Could you give me some guidance? How would I invoke the Gradle buildstep? How would I use the `gradle` symbol on the `ToolInstallation`?

PR #9, opened 2012-08-16
JENKINS-12769/JENKINS-14272 - specifying custom gradlew location

noamt:
Enable specification of a relative path to the gradlew executable within the workspace; supports also remote agents

PR #37, opened 2016-09-06
Make it possible to configure the wrapper location

wolfs:
Currently, it is only possible to configure the location of the wrapper in the root directory of the build or the root directory of the workspace checkout. It is desirable to configure it in other places as well. There are actually two other pull requests already implementing this:
- https://github.com/jenkinsci/gradle-plugin/pull/11
- https://github.com/jenkinsci/gradle-plugin/pull/9

This PR basically is a mix of these two PRs, adding tests on top of them and making it work with current master in a backwards compatible way.  The issues we need to solve are also written down in https://issues.jenkins-ci.org/browse/JENKINS-35029.

@gezerk : What do you think?


gezerk:
@wolfs this looks great.  I will build it and test it today.  I will also look at the failing merge build.

src/main/java/hudson/plugins/gradle/Gradle.java
```diff
@@ -190,9 +185,13 @@ private boolean performTask(boolean dryRun, AbstractBuild<?, ?> build, Launcher
         //Build arguments
         ArgumentListBuilder args = new ArgumentListBuilder();
         if (useWrapper) {
-            FilePath gradleWrapperFile = getGradleWrapperFile(build, launcher, normalizedRootBuildScriptDir);
+            FilePath gradleWrapperFile = getGradleWrapperFile(build, launcher, env, normalizedRootBuildScriptDir);
+            if( !gradleWrapperFile.exists() ) {
+                listener.fatalError("Unable to find Gradle Wrapper");
```
gezerk: It would be nice if this provided more detail about where it looked for the wrapper.

src/main/resources/hudson/plugins/gradle/Gradle/config.jelly
```diff
@@ -16,8 +16,8 @@
         <f:entry title="${%Make gradlew executable}" field="makeExecutable">
             <f:checkbox />
         </f:entry>
-        <f:entry title="${%From Root Build Script Dir}" field="fromRootBuildScriptDir">
-            <f:checkbox default="true"/>
+        <f:entry title="${%Wrapper location}" field="wrapperLocation"> <!-- Optional -->
```
gezerk: When I configure the job to use a wrapper it's not entirely clear where the wrapper comes from.  It would be nice if the plugin explained where it searches for the wrapper and that providing the location is only needed if the wrapper is not in one of the default locations.

wolfs:
@gezerk Thanks for the feedback! I tried to address it. Could you have another look?

gezerk:
Really appreciate that.  It looks great!  Want me to merge it once the tests pass or will you?

wolfs:
Feel free to merge it if you think this is a good solution.

gezerk:
The PR merge is failing looks like a problem starting a docker container.  I think it is safe to merge anyway.  I tried to rebuild it but couldn't figure out how to.  What do you think?

wolfs:
@gezerk Yes, it just failed on the new `Jenkinsfile` infrastructure. Luckily, the PR-builder job succeeded. I am confident that this is safe to merge.

syslogic:
`/var/lib/jenkins` seems to be hard-coded somewhere; my `/etc/sysconfig/jenkins` has defined `JENKINS_HOME="/home/jenkins"` and the wrapper seems to be found at location `${workspace}`... but when building, it nevertheless tries to access the default `$JENKINS_HOME`:

    [Gradle] - Launching build.
    [XYZ] $ /home/jenkins/workspace/XYZ/gradlew build
    Exception in thread "main" java.io.FileNotFoundException: /var/lib/jenkins/.gradle/wrapper/dists/gradle-4.1-all/bzyivzo6n839fup2jbap0tjew/gradle-4.1-all.zip.lck (No such file or directory)

the actual location should be: `/home/jenkins/workspace/XYZ/gradle/wrapper/gradle-wrapper.jar`

wolfs:
@syslogic The problem you are describing seems to be related to the android plugin. Please open an issue on the Google issue tracker to find out more. You can also open a new issue on https://issues.jenkins-ci.org if you think this is more related to the Jenkins Gradle plugin.

PR #38, opened 2016-09-06
Tighten up the Jenkinsfile a bit

rtyler:


Jenkinsfile
```diff
@@ -9,41 +9,44 @@ node {
...
   String jdktool = tool name: "jdk7", type: 'hudson.model.JDK'
 
   /* Set JAVA_HOME, and special PATH variables. */
   List javaEnv = [
-    "PATH+JDK=${jdktool}/bin", "JAVA_HOME=${jdktool}",
+    "PATH+JDK=${jdktool}/bin",
+    "JAVA_HOME=${jdktool}",
     // Additional variables needed by tests on machines
     // that don't have global git user.name and user.email configured.
-    'GIT_COMMITTER_EMAIL=me@hatescake.com','GIT_COMMITTER_NAME=Hates','GIT_AUTHOR_NAME=Cake','GIT_AUTHOR_EMAIL=hates@cake.com', 'LOGNAME=hatescake'
```
bitwiseman: I think these are specific to git-plugin testing.

rtyler: I also use them to deal with [Git inside of Docker](https://github.com/jenkins-infra/jenkins.io/blob/master/Jenkinsfile#L55-L62) :smiley_cat: 

I might change this to just use Docker instead of tools like it is now, since Docker allows for more developer flexibility here

wolfs:
Thanks Tyler for fixing the Jenkinsfile. I merged it and the build works like a charm. As I understand it I also could change the format of the Jenkinsfile to the new `pipeline` syntax, right?

rtyler:
@wolfs It depends on what you mean by "new syntax" :smile: 

If you're referring to @abayer's work on the Pipeline Model Definition plugin, that is not installed in any of our environments and I don't anticipate installing it until it's matured a bit more.

wolfs:
@rtyler Yeah, I was referring to @abayer's work on the Pipeline Definition plugin. Thanks for the clarification.

PR #42, opened 2017-02-28
[FIXED JENKINS-42381] Add PATH+GRADLE and buildEnvVars(...)

abayer:
[JENKINS-42381](https://issues.jenkins-ci.org/browse/JENKINS-42381)

This ensures that Gradle's bin directory ends up on the PATH when
using the Pipeline tool step. Note that Jenkins' environment variable
management will handle converting "/bin" on Windows if needed, we do
not need to do that explicitly.

cc @reviewbybees and @wolfs

### DIFF
--- a/src/test/groovy/hudson/plugins/gradle/GradlePluginIntegrationTest.groovy
+++ b/src/test/groovy/hudson/plugins/gradle/GradlePluginIntegrationTest.groovy
@@ -28,6 +28,7 @@ import com.gargoylesoftware.htmlunit.html.HtmlButton
 import com.gargoylesoftware.htmlunit.html.HtmlForm
 import com.gargoylesoftware.htmlunit.html.HtmlPage
 import com.google.common.base.Joiner
+import hudson.EnvVars
 import hudson.model.Cause
 import hudson.model.FreeStyleBuild
 import hudson.model.FreeStyleProject
@@ -270,6 +271,11 @@ task hello << { println 'Hello' }"""))
         def installation = installations[0]
         installation.name == 'myGradle'
         installation.home == '/tmp/foo'
+
+        def envVars = new EnvVars()
+        installation.buildEnvVars(envVars)
+        assert envVars.containsKey("PATH+GRADLE")
+        assert envVars.get("PATH+GRADLE") == installation.home + "/bin"
 
         // by default we should get the auto installer
         def props = installations[0].getProperties()
--- a/src/main/java/hudson/plugins/gradle/GradleInstallation.java
+++ b/src/main/java/hudson/plugins/gradle/GradleInstallation.java
@@ -63,6 +63,10 @@ public class GradleInstallation extends ToolInstallation
         return super.getHome();
     }
 
+    @Override
+    public void buildEnvVars(EnvVars env) {
+        env.put("PATH+GRADLE", getHome() + "/bin");
+    }
 
     public String getExecutable(Launcher launcher) throws IOException, InterruptedException {
         return launcher.getChannel().call(new MasterToSlaveCallable<String, IOException>() {


wolfs:
Thanks for your pull request.

Could you add an integration test exposing the problem you fix here?

src/test/groovy/hudson/plugins/gradle/GradlePluginIntegrationTest.groovy
```diff
@@ -271,6 +272,11 @@ task hello << { println 'Hello' }"""))
         installation.name == 'myGradle'
         installation.home == '/tmp/foo'
 
+        def envVars = new EnvVars()
+        installation.buildEnvVars(envVars)
+        assert envVars.containsKey("PATH+GRADLE")
+        assert envVars.get("PATH+GRADLE") == installation.home + "/bin"
```
wolfs: By an integration test I rather meant that the path to `GRADLE` is really added to the path and then we can execute gradle. Preferable in a pipeline job. Is that possible?

jglick: This is a unit test. An integration test would actually demonstrate that `$PATH` is right.

abayer: Touché. Lemme figure out how I can test the `PATH` here.

abayer: Lemme see what I can do.

jglick: Since the stated use case is for Declarative, perhaps you should have that as a test dep and demonstrate that

```groovy
tools {…}
steps {
  echo "PATH=$PATH"
}
```

works.

abayer: Downside to that is that we'd have to bump the Gradle plugin's core dependency to 2.7.1 from its current 1.642.1.

abayer:
@wolfs @jglick The only way I can see to do an integration test is by bumping the core dependency to 2.7.1 so we can use Declarative as a test dependency. As I said over on the related Ant PR (https://github.com/jenkinsci/ant-plugin/pull/26#discussion_r103558686), I think that verifying that `PATH+GRADLE` gets added to the environment by `buildEnvVars` is sufficient - that's exactly the path that Declarative exercises (https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/steps/EnvVarsForToolStep.java#L130-L136) and passes onward to `withEnv`. But, your call(s).

wolfs:
@abayer Is it possible to build a plugin against one version of Jenkins and test against a few other versions? Could I just create a few test configurations containing different Jenkins versions and it would work?

abayer:
@wolfs I honestly don't know. =)

jglick:
> Is it possible to build a plugin against one version of Jenkins and test against a few other versions?

No. You can _run_ tests against whatever versions you like, by passing `-Djenkins.version=…` on the command line, but you cannot _write_ tests against any version other than your actual baseline. So I guess this is a dead end.

wolfs:
@abayer Well, looks like this is all the test coverage I get for now. Could you please add something to the release notes in the [README](https://github.com/jenkinsci/gradle-plugin/blob/master/README.md)?

wolfs:
This has been merged manually by https://github.com/jenkinsci/gradle-plugin/commit/e195cf311146ceefd48b98a77cf1699a51268865.

wolfs:
Thank you for your contribution @abayer!

PR #48, opened 2017-05-08
[JENKINS-44126] Include animal-sniffer and findbugs

varyvol:
[JENKINS-44126](https://issues.jenkins-ci.org/browse/JENKINS-44126)

Findbugs and animal-sniffer analyses are really useful for development.

- Execute animal sniffer when building.
- Execute Findbugs when building.
- Fix Findbugs errors.

@reviewbybees @wolfs

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -72,7 +72,6 @@ findbugs {
 }
 
 codenarc {
-  // Some errors are logged due to https://github.com/gradle/gradle/issues/1094, but Findbugs checks are working fine
   toolVersion '0.27.0'
   sourceSets = [sourceSets.test]
 }
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,7 @@ plugins {
   id "org.jenkins-ci.jpi" version "0.22.0"
   id 'ru.vyarus.animalsniffer' version '1.3.0'
   id 'findbugs'
+  id 'codenarc'
 }
 
 group = "org.jenkins-ci.plugins"
@@ -68,6 +69,13 @@ findbugs {
   // Some errors are logged due to https://github.com/gradle/gradle/issues/1094, but Findbugs checks are working fine
   toolVersion '3.0.0'
   ignoreFailures false
+  sourceSets = [sourceSets.main]
+}
+
+codenarc {
+  // Some errors are logged due to https://github.com/gradle/gradle/issues/1094, but Findbugs checks are working fine
+  toolVersion '0.27.0'
+  sourceSets = [sourceSets.test]
 }
 
 test {
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,7 @@
 plugins {
   id "org.jenkins-ci.jpi" version "0.22.0"
+  id 'ru.vyarus.animalsniffer' version '1.3.0'
+  id 'findbugs'
 }
 
 group = "org.jenkins-ci.plugins"
@@ -46,6 +48,8 @@ dependencies {
   compileOnly 'org.jenkins-ci:symbol-annotation:1.3'
   jenkinsPlugins 'org.jenkins-ci.plugins:structs:1.3@jar'
 
+  signature 'org.codehaus.mojo.signature:java16:1.1@signature'
+
   testCompile 'org.spockframework:spock-core:0.7-groovy-1.8'
   jenkinsTest 'org.jenkins-ci.main:jenkins-test-harness:2.8@jar'
 }
@@ -56,6 +60,16 @@ if (project.hasProperty("maxParallelForks")) {
   ext.maxParallelForks = 3
 }
 
+animalsniffer {
+  ignoreFailures true // https://github.com/gradle/gradle/issues/1305
+}
+
+findbugs {
+  // Some errors are logged due to https://github.com/gradle/gradle/issues/1094, but Findbugs checks are working fine
+  toolVersion '3.0.0'
+  ignoreFailures false
+
+}
 
 test {
   systemProperties["hudson.model.DownloadService.noSignatureCheck"] = "true"
--- a/build.gradle
+++ b/build.gradle
@@ -68,7 +68,6 @@ findbugs {
   // Some errors are logged due to https://github.com/gradle/gradle/issues/1094, but Findbugs checks are working fine
   toolVersion '3.0.0'
   ignoreFailures false
-
 }
 
 test {
--- a/build.gradle
+++ b/build.gradle
@@ -62,13 +62,12 @@ if (project.hasProperty("maxParallelForks")) {
 }
 
 animalsniffer {
-  ignoreFailures true // https://github.com/gradle/gradle/issues/1305
+  sourceSets = [sourceSets.main]
 }
 
 findbugs {
   // Some errors are logged due to https://github.com/gradle/gradle/issues/1094, but Findbugs checks are working fine
   toolVersion '3.0.0'
-  ignoreFailures false
   sourceSets = [sourceSets.main]
 }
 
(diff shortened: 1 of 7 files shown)


armfergom:
Changes look ok to me, but the build is failing. Why?

oleg-nenashev:
```
Condition not satisfied:
installers.size() == 1
|          |      |
|          5      false
[hudson.plugins.gradle.GradleInstaller@1d6968b9, hudson.tools.ZipExtractionInstaller@296bf56d, hudson.plugins.gradle.GradleInstaller@1de7bab6, hudson.tools.BatchCommandInstaller@23a36ce2, hudson.tools.CommandInstaller@75d9f48f]
Stacktrace
Condition not satisfied:
installers.size() == 1
|          |      |
|          5      false
[hudson.plugins.gradle.GradleInstaller@1d6968b9, hudson.tools.ZipExtractionInstaller@296bf56d, hudson.plugins.gradle.GradleInstaller@1de7bab6, hudson.tools.BatchCommandInstaller@23a36ce2, hudson.tools.CommandInstaller@75d9f48f]
	at hudson.plugins.gradle.GradlePluginIntegrationTest.installationConfigured(GradlePluginIntegrationTest.groovy:229)
	at hudson.plugins.gradle.GradlePluginIntegrationTest.add Gradle installation(GradlePluginIntegrationTest.groovy:210)
```

src/test/groovy/hudson/plugins/gradle/PropertyPassingIntegrationTest.groovy
```diff
@@ -1,5 +1,6 @@
 package hudson.plugins.gradle
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings
```
wolfs: Could you please not use Findbugs for Groovy integration tests? Add Codenarc instead!

wolfs:
@oleg-nenashev This functionality should be add to the gradle-jpi-plugin, so all Jenkins plugins authors using Gradle can benefit from them. This can be either as
- different plugins within the `gradle-jpi-plugin` repository
- having these things configurable via the build script.

build.gradle
```diff
@@ -56,6 +60,15 @@ if (project.hasProperty("maxParallelForks")) {
...
 }
 
+animalsniffer {
+  ignoreFailures true // https://github.com/gradle/gradle/issues/1305
+}
+
+findbugs {
+  // Some errors are logged due to https://github.com/gradle/gradle/issues/1094, but Findbugs checks are working fine
+  toolVersion '3.0.0'
+  ignoreFailures false
```
wolfs: Isn't this the default?

wolfs:
Thank you for your PR @varyvol. I added some comments to the code. Please address those to get this PR merged.

src/test/groovy/hudson/plugins/gradle/PropertyPassingIntegrationTest.groovy
varyvol: You mean for this specific tests? Or for the whole project?

build.gradle
```diff
@@ -56,6 +60,15 @@ if (project.hasProperty("maxParallelForks")) {
   ext.maxParallelForks = 3
 }
 
+animalsniffer {
+  ignoreFailures true // https://github.com/gradle/gradle/issues/1305
```
varyvol: It's failing for me with the version specified in the wrapper (3.4.1) and actually even with 3.5.

build.gradle
varyvol: No, the build fails if you don't specifically set it not to.

src/test/groovy/hudson/plugins/gradle/PropertyPassingIntegrationTest.groovy
wolfs: Only use Findbugs on Java sources for the whole project please. Currently, the main sources are all Java and the test sources are all Groovy.

varyvol: Will do!

varyvol:
@wolfs no worries. I will try to work on your comments so that we can push this forward.

@armfergom it's working for me locally, and I don't think the changes could affect that part. Where did you see that log @oleg-nenashev? I am not able to find it in the build log.

varyvol:
@wolfs I have done changes based on your comments.

@armfergom @oleg-nenashev the build was successful now, I think it was something environmental.

wolfs:
@varyvol Thank you for addressing my comments. Findbugs and codenarc look good now.
Regarding animalsniffer: I digged a little bit deeper and the error persists with Gradle 3.3. This means that the error is not on the Gradle side but it is a error reported by animalsniffer. The dependency which causes the error is actually pulled in via the jenkins core dependency (see [here](https://scans.gradle.com/s/ldwtqgbbj5ttk/dependencies?dependencies=icu&focusedDependency=WzAsMSwxMTUsWzAsMCxbMF1dXQ%3D%3D&focusedDependencyView=dependents&search)). How is this problem solved for other plugins or the Jenkins build itself? Is there some ignoring in place?
I would not add a static analyzer checking tool just to ignore failures from it.

build.gradle
```diff
@@ -56,6 +61,22 @@ if (project.hasProperty("maxParallelForks")) {
   ext.maxParallelForks = 3
 }
 
+animalsniffer {
```
wolfs: Could we restrict animalsniffer to the main sourceset, too?

build.gradle
varyvol: I was actually wrong, so it can be removed.

build.gradle
varyvol: Sure.

varyvol:
@wolfs it does not seem to provoke any issue with Maven, and I have not found any other Gradle plugin that includes animal sniffer. I have excluded the dependency, which avoids the issue.

build.gradle
```diff
@@ -56,6 +61,26 @@ if (project.hasProperty("maxParallelForks")) {
...
+
+codenarc {
+  // Some errors are logged due to https://github.com/gradle/gradle/issues/1094, but Findbugs checks are working fine
+  toolVersion '0.27.0'
+  sourceSets = [sourceSets.test]
+}
+
+configurations {
+  // https://github.com/gradle/gradle/issues/1305
+  compile.exclude group: 'com.ibm.icu', module: 'icu4j'
```
wolfs: Are you sure I don't need this dependency for running Jenkins? I am also somewhat afraid the generated HPI file could by wrong by doing this. What does the Maven build do to support animal sniffer?

varyvol: Nothing extraordinary. Just declares the dependency and uses this configuration:
`<plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>animal-sniffer-maven-plugin</artifactId>
        <executions>
          <execution>
            <goals>
              <goal>check</goal>
            </goals>
            <id>check</id>
          </execution>
        </executions>
        <configuration>
          <signature>
            <groupId>org.codehaus.mojo.signature</groupId>
            <artifactId>java1${java.level}</artifactId>
          </signature>
        </configuration>`

As far as I know, the HPI plugin does not include either the core (as it is a provided dependency) nor other plugins dependencies (as they are handled by the plugin manager).

build.gradle
```diff
@@ -56,6 +61,26 @@ if (project.hasProperty("maxParallelForks")) {
...
+}
+
+findbugs {
+  // Some errors are logged due to https://github.com/gradle/gradle/issues/1094, but Findbugs checks are working fine
+  toolVersion '3.0.0'
+  sourceSets = [sourceSets.main]
+}
+
+codenarc {
+  // Some errors are logged due to https://github.com/gradle/gradle/issues/1094, but Findbugs checks are working fine
```
jglick: copy-paste error

varyvol: Thanks. Changed.

wolfs:
@varyvol Thanks for your changes. I polished it a little bit more on `master` addressing some of my comments: https://github.com/jenkinsci/gradle-plugin/commit/fac55cf5f68035a1685ce2162f89578c20cb2122. Please have a look if you agree.

varyvol:
@wolfs they look great to me, only maybe the comment in the configurations section should use the word module/dependency instead of class the first time.

Thanks for reviewing and working into this.

PR #51, opened 2017-06-28
Build variables that are empty are removed

shanman190:
By utilizing the EnvVars.overrideAll(Map) method, it internally removes keys where the value is null or an empty string. In some cases, users that I'm supporting have optional build parameters being utilized. This results in those parameters not being rewritten to empty properties as expected and the properties instead have the jenkins replacement string in them instead.

Example (tasks block):
```
:sometask -PPROPERTY=${OPTIONAL_PROPERTY}
```

If OPTIONAL_PROPERTY is a Jenkins build variable and it is left blank, we end up getting an incorrect value.

As workaround, we can remove the property from the tasks block itself and use the pass as gradle properties checkbox.

### DIFF
--- a/src/main/java/hudson/plugins/gradle/Gradle.java
+++ b/src/main/java/hudson/plugins/gradle/Gradle.java
@@ -228,7 +228,6 @@ public class Gradle extends Builder implements DryRun {
         gradleLogger.info("Launching build.");
 
         EnvVars env = build.getEnvironment(listener);
-        env.overrideAll(build.getBuildVariables());
         final VariableResolver<String> resolver = new VariableResolver.ByMap<>(env);
 
         //Switches


wolfs:
@shanman190 Thank you very much for your pull request!
Could you also provide an integration test which shows the problem you are trying to solve?
Could you be a little bit more specific in "If OPTIONAL_PROPERTY is a Jenkins build variable and it is left blank, we end up getting an incorrect value."? What is the incorrect value and in what sense is it incorrect?
In removing the `overrideAll` line, do we also loose some other build variables which could be used in variable expansion?

wolfs:
I guess this PR would fix [JENKINS-45300](https://issues.jenkins-ci.org/browse/JENKINS-45300), right?

shanman190:
@wolfs, Yep that ticket is exactly the problem. 

By default EnvVars has already been populated with all of the environment variables as well as all of the build parameters. The call to `overrideAll` internally removes any property that evaluates to null or empty. Then when you later make the call to expand the values, the placeholder is left as is.

I can try to get an test for it written up to make sure that the problem doesn't regress.

wolfs:
@shanman190 I opened #53 with tests and the fix for this problem. Please have a look and try out if it works for you. I will close this PR, please comment on #53 .

PR #53, opened 2017-08-13
Don't remove empty build vars

wolfs:
By utilizing the EnvVars.overrideAll(Map) method, it internally removes keys where the value is null or an empty string. This results in those parameters not being rewritten to empty properties as expected and the properties instead have the jenkins replacement string in them instead.

[FIXES JENKINS-45300]

### DIFF
--- a/src/main/java/hudson/plugins/gradle/Gradle.java
+++ b/src/main/java/hudson/plugins/gradle/Gradle.java
@@ -170,8 +170,8 @@ public class Gradle extends Builder implements DryRun {
     public String getProjectProperties() {
         return projectProperties;
     }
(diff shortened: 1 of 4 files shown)


wolfs:
@shanman190: Could you try out this change and verify it fixes your problems ([JENKINS-45300](https://issues.jenkins-ci.org/browse/JENKINS-45300))

shanman190:
@wolfs, looks like it should work perfectly. I'll try to test out this PR either this evening or tomorrow.

shanman190:
@wolfs, was successfully able to test and verify that the issue is in fact resolved. Thanks!

PR #55, opened 2018-05-02
[JENKINS-51062] JEP-200 issue with plugin Gradle Plugin

fcojfernandez:
See [JENKINS-51062](https://issues.jenkins-ci.org/browse/JENKINS-51062)

- Add the hudson.remoting.ClassFilter since all the involved classes seems to be fine (non transient field are String)
- Upgrade `org.jenkins-ci.jpi` to version 0.24.0 to avoid the use of war-for-test when checking the compatibility with newer core versions.

@reviewbybees @jglick @oleg-nenashev @wolfs

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 import java.util.zip.ZipFile
 
 plugins {
-  id "org.jenkins-ci.jpi" version "0.22.0"
+  id "org.jenkins-ci.jpi" version "0.24.0"
   id 'ru.vyarus.animalsniffer' version '1.3.0'
   id 'findbugs'
   id 'codenarc'
--- a/build.gradle
+++ b/build.gradle
@@ -62,7 +62,8 @@ dependencies {
   signature 'org.codehaus.mojo.signature:java17:1.0@signature'
 
   testCompile 'org.spockframework:spock-core:0.7-groovy-1.8'
-  jenkinsTest 'org.jenkins-ci.main:jenkins-test-harness:2.8@jar'
+  jenkinsTest 'org.jenkins-ci.main:jenkins-test-harness:2.38@jar'
+  compile 'junit:junit:4.12'
 }
 
 if (project.hasProperty("maxParallelForks")) {
--- a/src/main/resources/META-INF/hudson.remoting.ClassFilter
+++ b/src/main/resources/META-INF/hudson.remoting.ClassFilter
@@ -1,6 +1,0 @@
-hudson.plugins.gradle.BuildScanAction
-hudson.plugins.gradle.Gradle
-hudson.plugins.gradle.Gradle$DescriptorImpl
-hudson.plugins.gradle.GradleInstallation
-hudson.plugins.gradle.GradleInstallation$DescriptorImpl
-hudson.plugins.gradle.GradleInstaller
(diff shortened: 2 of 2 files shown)


fcojfernandez:
@oleg-nenashev I'm afraid I tried but i was not success

oleg-nenashev:
We had a screenshare with @fcojfernandez . The root cause is likely using an obsolete JTH version which does not do classloading properly. Tooling needs to be updated, there should be no issue after that

fcojfernandez:
@oleg-nenashev after updating test-harness and adding the correct junit version, the issue is still reproducing when jenkins version is forced to `2.107`. I'm still not able to reproduce it locally. Any other idea?

jglick:
So far I am unable to run with your patches plus

```diff
diff --git a/build.gradle b/build.gradle
index 9a3dba5..9b96bc9 100644
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ if (ciBuild) {
 
 jenkinsPlugin {
   // Version of Jenkins core this plugin depends on.
-  coreVersion = "1.642.1"
+  coreVersion = "2.107.2"
 
   // Human-readable name of plugin.
   displayName = "Gradle Plugin"
```

since

```bash
./gradlew build -x findbugsMain
```

fails with

```
Could not instantiate global transform class org.spockframework.compiler.SpockTransform specified at jar:file:…/.gradle/caches/modules-2/files-2.1/org.spockframework/spock-core/0.7-groovy-1.8/3a677d19e8d3acf3bd296c4023356256d55da5a3/spock-core-0.7-groovy-1.8.jar!/META-INF/services/org.codehaus.groovy.transform.ASTTransformation  because of exception org.spockframework.util.IncompatibleGroovyVersionException: The Spock compiler plugin cannot execute because Spock 0.7.0-groovy-1.8 is not compatible with Groovy 2.4.11. For more information, see http://versioninfo.spockframework.org
```

Again, please specify exactly what you were doing to reproduce the test failure.

jglick:
Assuming https://github.com/jenkinsci/jenkins/pull/3420 deploys successfully, try

```diff
diff --git a/build.gradle b/build.gradle
index 9a3dba5..197ff08 100644
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ if (ciBuild) {
 
 jenkinsPlugin {
   // Version of Jenkins core this plugin depends on.
-  coreVersion = "1.642.1"
+  coreVersion = "2.120-rc15714.b14ba46a141b"
 
   // Human-readable name of plugin.
   displayName = "Gradle Plugin"
@@ -61,7 +61,7 @@ dependencies {
 
   signature 'org.codehaus.mojo.signature:java17:1.0@signature'
 
-  testCompile 'org.spockframework:spock-core:0.7-groovy-1.8'
+  testCompile 'org.spockframework:spock-core:1.0-groovy-2.4'
   jenkinsTest 'org.jenkins-ci.main:jenkins-test-harness:2.38@jar'
   compile 'junit:junit:4.12'
 }
```

jglick:
Actually it _was_ deployed successfully; it is just that you are looking in the wrong repo. It is in https://repo.jenkins-ci.org/incrementals/ instead. AFAICT the `gradle-plugin` sources do not mention Maven repos so you would need to configure `~/.m2/settings.xml`. Maven-based Jenkins plugins specify `<repositories>`s explicitly, and with a new parent POM could be configured to `-Pconsume-incrementals` to pick up this repo.

fcojfernandez:
This PR does not make sense now, so I will close it.

Thanks @jglick and @oleg-nenashev

PR #59, opened 2018-10-03
Convert Plugin to be compatible with Jenkins pipeline

funtastix:
- DryRun was removed for this purpose and SimpleBuildStep implemented instead
- Simple usage example: "gradle tasks: 'clean assembleDebug', useWrapper: true"
- Works with Snippet Generator (which adds additional/unnecessary parameters)
- This enables gradle (wrappers) to be used easily cross-platform without scripts needed

### DIFF
--- a/src/main/java/hudson/plugins/gradle/Gradle.java
+++ b/src/main/java/hudson/plugins/gradle/Gradle.java
@@ -2,26 +2,28 @@ package hudson.plugins.gradle;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
+
 import hudson.CopyOnWrite;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
 import hudson.Util;
-import hudson.model.AbstractBuild;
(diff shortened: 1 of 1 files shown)


jvz:
Thanks for the contribution! Can you fix the tests?

funtastix:
I will look into them. There seems to be additional tests running as all test had passed on my build machine.

jvz:
The test failures appear to be on Windows only, so that could be part of the problem.

funtastix:
This is what is confusing as I build on windows. I will hopefully figure it out.

funtastix:
@jvz I tracked the failures down to a parameter passing ending in '\'. I don't understand what this needs to look like and how my changes would have affected those  tests. The parameter encoding is done via Jenkins until classes. I will try to get a comparison with and without my commits. It may be a good idea to add a new parameter test: Param="XYZ\"

jvz:
Almost sounds like the test failure might be unrelated then?

wolfs:
@funtastix I'll be closing the PR, given that there hasn't been any activity. Is the PR still interesting? If yes, please open a new one based on current `main`. Thank you!

PR #52, opened 2017-06-28
[JENKINS-44834] Partial pipeline support 

alexbrjo:
[JENKINS-44834](https://issues.jenkins-ci.org/browse/JENKINS-44834) Adds a `withGradle` pipeline step to set up the Gradle environment and annotates the build output. Provides Gradle and JDK tool selection from pipeline.

- [x] Waiting on [gradle-jpi-plugin#87](https://github.com/jenkinsci/gradle-jpi-plugin/pull/87) to be merged.
- [X] Tests rely on a local Gradle installation. Waiting on a [jenkins-test-harness-tools](https://github.com/jenkinsci/jenkins-test-harness-tools) release.

PR #60, opened 2018-11-22
[JENKINS-44834]: finished "withGradle" step

soenkekueper:
I've just done some works on the initial fork by @alexbrjo to use the org.jenkins-ci.jpi gradle plugin.
see https://github.com/jenkinsci/gradle-plugin/pull/52 for furhter information.

Along with some findbugs fixes and env var fix on windows this would be really nice to get gradle a little bit more included in the jenkins pipeline feature.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -70,6 +70,7 @@ sourceCompatibility = '1.7'
 
 dependencies {
   compile 'org.jenkins-ci.lib:dry-run-lib:0.1'
+  compile 'junit:junit:4.12' // TODO: remove. this should be included in jenkins-test-harness
   signature 'org.codehaus.mojo.signature:java17:1.0@signature'
 
   testCompile 'org.spockframework:spock-core:0.7-groovy-1.8'
--- a/build.gradle
+++ b/build.gradle
@@ -69,27 +69,20 @@ jenkinsPlugin {
 sourceCompatibility = '1.7'
 
 dependencies {
-  compile 'org.jenkins-ci.lib:dry-run-lib:0.1'
-  compile 'junit:junit:4.12' // TODO: remove. this should be included in jenkins-test-harness
+  jenkinsPlugins 'org.jenkins-ci.plugins:dry-run:0.9'
+  jenkinsPlugins 'org.jenkins-ci.plugins.workflow:workflow-step-api:2.11'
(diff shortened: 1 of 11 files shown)


soenkekueper:
Hey,

we are using gradle in a jenkins pipeline and had to do a lot of manually work to get it running.
Than i've discovered the branch and pull request #52 by @alexbrjo and thought that would make it a bit easier.
Due there was only little work to do, i've finished this request.

After this change, get a gradle running in an jenkinsfile is a little bit more easier - as you can use the gradle and jdk installation configured in jenkins, and - for usability and readability of the logs - the gradle console annotator is used. So you can see (as in regular builds or with ant) the tasks that have been executed and jump directly to them.

But there is one more problem which - i've also ran into when creating this pull request - you have to build your gradle command line manually and you have to care about platform specifics (as you can see in the tests you must always check if you are running on unix or windows and call sh or bat).

Due the logic of how to build the gradle command and to create an platform independent gradle step - i've afterwards started working on an simple "real" gradle step so you don't have to do this any more.

To wrap up:
You are now able to run a pipline like this

    node {
      withGradle(gradleName:'gradle-4.10') {
        if (isUnix()) {
      	sh 'gradle'
        } else {
          bat 'gradle build'
        }
      }
    }

and you don't have to modify your PATH and JAVA_HOME (which in many cases causes other problems) and you have a little bit more readability of the logs.

But what im looking forward (and working on) is to have a real gradle step to write a jenkins file like this,
which could aviod the problems above.

    node {
        gradle gradleName: 'gradle-4.10', tasks: 'cleanTest build'
    }

DJViking:
What needs to be done to get this finished and merged?

I have one comment on the withGradle step. The property naming of gradleName and jdkName
Why not just call it gradle and jdk? Then it would be more aligned with withMaven step naming of these same properties.

```
  withGradle(gradleName:'gradle-4.10', jdkName: 'jdk-11') {}
```

```
  withGradle(gradle:'gradle-4.10', jdk: 'jdk-11') {}
```

```
  withMaven(maven:'maven-3.5', jdk: 'jdk-11') {}
```

wolfs:
I'll be closing the PR, given that there hasn't been any recent activity. Is the PR still interesting? If yes, please open a new one based on current `main`. Thank you!

PR #61, opened 2018-11-29
Implemented gradle pipeline step

soenkekueper:
Hey,
i've now put some more work on a real gradle pipeline step. This makes it **much** easier and comfortable running a gradle build in an jenkins pieline.

You are now able to run a gradle build using a predefined gradle with just this little pipeline
`node{
    gradle gradleName: 'gradle-4.10.2', tasks: 'clean build'
}`

I've implemented this in our test environment and saved a lot of code in our pipeline script for handling environment variables like PATH and JAVA_HOME catching stderr of gradle logs and so on.

It would be great to get this in so we could use this in our productive environment.

Basically i've moved the main logic of executing gradle into GradleExecution.java so it can now be used by pipeline and freestyle builds as well.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 import java.util.zip.ZipFile
 
 plugins {
-  id "org.jenkins-ci.jpi" version "0.22.0"
+  id 'org.jenkins-ci.jpi' version '0.23.1'
   id 'ru.vyarus.animalsniffer' version '1.3.0'
   id 'findbugs'
   id 'codenarc'
@@ -18,7 +18,7 @@ if (ciBuild) {
 
 jenkinsPlugin {
   // Version of Jenkins core this plugin depends on.
-  coreVersion = "1.642.1"
+  coreVersion = "1.642.3"
 
   // Human-readable name of plugin.
   displayName = "Gradle Plugin"
@@ -55,14 +55,21 @@ jenkinsPlugin {
 sourceCompatibility = '1.7'
 
 dependencies {
-  compile 'org.jenkins-ci.lib:dry-run-lib:0.1'
-  compileOnly 'org.jenkins-ci:symbol-annotation:1.3'
-  jenkinsPlugins 'org.jenkins-ci.plugins:structs:1.3@jar'
+  jenkinsPlugins 'org.jenkins-ci.plugins:dry-run:0.9'
+  jenkinsPlugins 'org.jenkins-ci.plugins.workflow:workflow-step-api:2.11'
+  jenkinsPlugins 'org.jenkins-ci.plugins:structs:1.7'
+  jenkinsPlugins 'org.jenkins-ci.plugins.workflow:workflow-support:1.14'
 
   signature 'org.codehaus.mojo.signature:java17:1.0@signature'
 
   testCompile 'org.spockframework:spock-core:0.7-groovy-1.8'
-  jenkinsTest 'org.jenkins-ci.main:jenkins-test-harness:2.8@jar'
+  testCompile 'org.jenkins-ci.main:jenkins-test-harness:2.22@jar'
+  testCompile 'org.jenkins-ci.main:jenkins-test-harness-tools:2.2'
+
+  jenkinsTest 'org.jenkins-ci.plugins.workflow:workflow-cps:2.32'
+  jenkinsTest 'org.jenkins-ci.plugins.workflow:workflow-job:2.11'
+  jenkinsTest 'org.jenkins-ci.plugins.workflow:workflow-basic-steps:2.5'
+  jenkinsTest 'org.jenkins-ci.plugins.workflow:workflow-durable-task-step:2.9'
 }
 
 if (project.hasProperty("maxParallelForks")) {
@@ -79,6 +86,9 @@ configurations {
   animalsnifferCompileClasspath.extendsFrom compile
   // We need to exclude this dependency from animalsniffer since contains an invalid class
   animalsnifferCompileClasspath.exclude group: 'com.ibm.icu', module: 'icu4j'
+  // junit-dep:4.10 cannot be replaced by junit:4.12 since the artifactIds are different
+  // gibut is causing binary compatibility issues. Exclude it completely from classpath
+  runtime.exclude group: 'junit', module: 'junit-dep'
 }
 
 findbugs {
--- a/src/main/java/hudson/plugins/gradle/Gradle.java
+++ b/src/main/java/hudson/plugins/gradle/Gradle.java
@@ -1,41 +1,34 @@
 package hudson.plugins.gradle;
 
-import com.google.common.base.Joiner;
-import com.google.common.collect.ImmutableList;
-import hudson.CopyOnWrite;
-import hudson.EnvVars;
-import hudson.Extension;
-import hudson.FilePath;
-import hudson.Launcher;
-import hudson.Util;
-import hudson.model.AbstractBuild;
-import hudson.model.AbstractProject;
-import hudson.model.BuildListener;
-import hudson.model.Computer;
-import hudson.model.Node;
-import hudson.model.Result;
-import hudson.tasks.BuildStepDescriptor;
-import hudson.tasks.Builder;
-import hudson.tools.ToolInstallation;
-import hudson.util.ArgumentListBuilder;
-import hudson.util.VariableResolver;
-import org.apache.commons.lang.StringUtils;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Map;
+
 import org.jenkinsci.Symbol;
 import org.jenkinsci.lib.dryrun.DryRun;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.DataBoundSetter;
 
-import java.io.IOException;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import hudson.CopyOnWrite;
+import hudson.EnvVars;
+import hudson.Extension;
+import hudson.Launcher;
+import hudson.Util;
+import hudson.model.AbstractBuild;
+import hudson.model.AbstractProject;
+import hudson.model.BuildListener;
+import hudson.tasks.BuildStepDescriptor;
+import hudson.tasks.Builder;
+import hudson.tools.ToolInstallation;
+import hudson.util.VariableResolver;
 
 
 /**
+ * {@link Builder} for Freestyle-Builds to execute Gradle.
+ * 
  * @author Gregory Boissinot
  */
-public class Gradle extends Builder implements DryRun {
+public class Gradle extends Builder implements DryRun, GradleInstallationProvider {
 
     private String switches;
     private String tasks;
@@ -54,11 +47,22 @@ public class Gradle extends Builder implements DryRun {
     private boolean passAllAsProjectProperties;
 
     private transient boolean fromRootBuildScriptDir;
+    
+    private String logEncoding;
 
     @DataBoundConstructor
     public Gradle() {
     }
 
+    @DataBoundSetter
+    public void setLogEncoding(String logEncoding) {
+        this.logEncoding = logEncoding;
+    }
+    
+    public String getLogEncoding() {
+        return logEncoding;
+    }
+    
     @SuppressWarnings("unused")
     public String getSwitches() {
         return switches;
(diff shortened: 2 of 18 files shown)


src/main/java/hudson/plugins/gradle/GradleExecution.java
```diff
@@ -0,0 +1,332 @@
...
+import hudson.util.ArgumentListBuilder;
+import hudson.util.VariableResolver;
+
+/**
+ * Helper for execution of an Gradle build.
+ * Encapsulates the real gradle execution independent of build type.
+ */
+public final class GradleExecution {
+
+    private String switches;
```
jvz: Looks like all these fields can be final?

src/main/java/hudson/plugins/gradle/Gradle.java
```diff
@@ -185,35 +189,6 @@ public void setPassAllAsProjectProperties(boolean passAllAsProjectProperties) {
         this.passAllAsProjectProperties = passAllAsProjectProperties;
     }
 
-    public GradleInstallation getGradle() {
```
jvz: Removing this here won't break any API compatibility, will it?

src/main/java/hudson/plugins/gradle/GradleInstallation.java
```diff
@@ -14,23 +24,18 @@
...
+ * 
  * @author Gregory Boissinot
  */
-public class GradleInstallation extends ToolInstallation
-        implements EnvironmentSpecific<GradleInstallation>, NodeSpecific<GradleInstallation>, Serializable {
+public class GradleInstallation 
+        extends ToolInstallation
+        implements EnvironmentSpecific<GradleInstallation>, NodeSpecific<GradleInstallation> {
 
+    private static final long serialVersionUID = -1875114350592080381L;
```
jvz: Why is this added if you removed `Serializable`? If it's still supposed to be serializable, you can keep the interface in the class declaration for explicitness.

src/main/java/hudson/plugins/gradle/GradleInstallationProvider.java
```diff
@@ -0,0 +1,15 @@
...
+ * Interface for providing {@link GradleInstallation} independent of build type.
+ * 
+ * @author Sönke Küper
+ */
+public interface GradleInstallationProvider {
+
+    /**
+     * Provides all available {@link GradleInstallation}.
+     */
+    public abstract GradleInstallation[] getInstallations();
```
jvz:
```suggestion
    GradleInstallation[] getInstallations();
```

`public` and `abstract` are implied by being in an interface.

src/main/java/hudson/plugins/gradle/GradleStep.java
```diff
@@ -0,0 +1,321 @@
...
+import hudson.tools.ToolInstallation;
+import hudson.util.VariableResolver;
+import jenkins.model.Jenkins;
+
+/**
+ * The Gradle-Step for a jenkins pipeline.
+ * 
+ * Only contains the configuration, the execution is delegated to an {@link GradleExecution}.
+ * 
+ * @author Sönke Küper
```
jvz: This doesn't match with the copyright header you've added. Can you fix one or the other as appropriate?

src/main/java/hudson/plugins/gradle/GradleExecution.java
soenkekueper: Yes, you are right!
I'll fix.

src/main/java/hudson/plugins/gradle/Gradle.java
soenkekueper: This was only used by the config.jelly with was also adjusted to now use getGradleName insted, so this should not be an problem.

src/main/java/hudson/plugins/gradle/GradleInstallation.java
soenkekueper: My IDE (eclipse) told me so ;) as the superclass ToolInstallation already implements Serializable i removed the duplicate specification. Is there any guideline concerning duplicate interface specifications?

src/main/java/hudson/plugins/gradle/GradleStep.java
soenkekueper: Yes, thanks for the hint.

src/main/java/hudson/plugins/gradle/Gradle.java
jvz: Ah ok, no problem there, then!

src/main/java/hudson/plugins/gradle/GradleInstallation.java
jvz: No guidelines that I can remember. It's more of a Java style thing in general.

soenkekueper: Okay you are right. i've added the Serializable interface!

src/main/java/hudson/plugins/gradle/GradleInstallationProvider.java
```diff
@@ -0,0 +1,15 @@
+package hudson.plugins.gradle;
+
+/**
+ * Interface for providing {@link GradleInstallation} independent of build type.
+ * 
+ * @author Sönke Küper
```
wolfs: Let's not use `@autor` tags. Git should be enough.

src/main/java/hudson/plugins/gradle/GradleInstallation.java
```diff
@@ -68,28 +73,46 @@ public void buildEnvVars(EnvVars env) {
...
-            public String call() throws IOException {
-                File exe = getExeFile();
-                if (exe.exists()) {
-                    return exe.getPath();
-                }
-                return null;
+    public String getExecutable(final Launcher launcher) throws IOException, InterruptedException {
+        final File exe = launcher.getChannel().call(new MasterToSlaveCallable<File, IOException>() {
+
+            private static final long serialVersionUID = 5572701806697595613L;
```
wolfs: What is the `serialVersionUID` for? IIUC, Jenkins uses XStream to serialize/deserialize data, not Java serialization. is this required?

src/main/java/hudson/plugins/gradle/GradleInstallationProvider.java
soenkekueper: Okay, i've just done it like in the other files around.. ;)

src/main/java/hudson/plugins/gradle/GradleInstallation.java
soenkekueper: Yes you are right, it's not needed here. Just a habit from my daily work. I'll remove it.

soenkekueper:
I've installed this in our test environment and it is running very well and has a lot of advantages instead of running a gradle by executing a shell or bat file. It makes it more comfortable, easier and safer to run a gradle build, because you don't have to set any environment variables in dependency to the node you are currently running on.

As i was looking for a solution to our problems i found the ticket JENKINS-44834 and existing pull request. So i first decided to finish the work started there leading to #60. Due the changes made there were prerequisite to implement a real step i've continued the work to implement this step. So in conclusion with this PR the PR #60 is obsolete.

build.gradle
```diff
@@ -55,14 +55,21 @@ jenkinsPlugin {
 sourceCompatibility = '1.7'
 
 dependencies {
-  compile 'org.jenkins-ci.lib:dry-run-lib:0.1'
-  compileOnly 'org.jenkins-ci:symbol-annotation:1.3'
-  jenkinsPlugins 'org.jenkins-ci.plugins:structs:1.3@jar'
+  jenkinsPlugins 'org.jenkins-ci.plugins:dry-run:0.9'
+  jenkinsPlugins 'org.jenkins-ci.plugins.workflow:workflow-step-api:2.11'
```
wolfs: I guess `workflow-step-api` and `worflow-support` should be in `optional JenkinsPlugins?

build.gradle
```diff
@@ -55,14 +55,21 @@ jenkinsPlugin {
...
 
   testCompile 'org.spockframework:spock-core:0.7-groovy-1.8'
-  jenkinsTest 'org.jenkins-ci.main:jenkins-test-harness:2.8@jar'
+  testCompile 'org.jenkins-ci.main:jenkins-test-harness:2.22@jar'
+  testCompile 'org.jenkins-ci.main:jenkins-test-harness-tools:2.2'
+
+  jenkinsTest 'org.jenkins-ci.plugins.workflow:workflow-cps:2.32'
+  jenkinsTest 'org.jenkins-ci.plugins.workflow:workflow-job:2.11'
+  jenkinsTest 'org.jenkins-ci.plugins.workflow:workflow-basic-steps:2.5'
+  jenkinsTest 'org.jenkins-ci.plugins.workflow:workflow-durable-task-step:2.9'
```
wolfs: I see that you depend on `durable-task-step`. Is this used somewhere? I would love the Gradle step to be a durable task step, i.e. when the master loses the connection to the `agent`, the build will continue and the master will re-connect to the agent and resume the operation for where it stopped.

Does this work with the current implementation? Or maybe the step should only support a nice API and the n run an `sh`/`bat` command under the covers.

src/main/java/hudson/plugins/gradle/GradleExecution.java
```diff
@@ -0,0 +1,332 @@
...
+                break;
+            }
+        }
+        if (gradleWrapperFile == null) {
+            listener.fatalError("The Gradle wrapper has not been found in these directories: %s", Joiner.on(", ").join(possibleWrapperLocations));
+        }
+        return gradleWrapperFile;
+    }
+
+    private FilePath getNormalizedRootBuildScriptDir(Run<?, ?> build, VariableResolver<String> resolver, FilePath moduleRoot) {
```
wolfs: `build` is not used anymore.

src/main/java/hudson/plugins/gradle/GradleExecution.java
```diff
@@ -0,0 +1,332 @@
...
+        normalizedArgs = replaceWhitespaceBySpace(normalizedArgs);
+        normalizedArgs = Util.replaceMacro(normalizedArgs, resolver);
+        return normalizedArgs;
+    }
+
+    private static String replaceWhitespaceBySpace(String argument) {
+        return argument.replaceAll("[\t\r\n]+", " ");
+    }
+
+    private List<FilePath> getPossibleWrapperLocations(Run<?, ?> build, Launcher launcher, VariableResolver<String> resolver, FilePath normalizedRootBuildScriptDir, FilePath moduleRoot) throws IOException, InterruptedException {
```
wolfs: `build` and `launcher` are not used anymore.

src/main/java/hudson/plugins/gradle/GradleInstallation.java
```diff
@@ -68,28 +72,42 @@ public void buildEnvVars(EnvVars env) {
...
+    public String getExecutable(final Launcher launcher) throws IOException, InterruptedException {
+        final File exe = launcher.getChannel().call(new MasterToSlaveCallable<File, IOException>() {
+
+            @Override
+            public File call() throws IOException {
+                return getExeFile();
+            }
+        });
+
+        final boolean existis = launcher.getChannel().call(new MasterToSlaveCallable<Boolean, IOException>() {
```
wolfs: Typo: should be `exists`.

build.gradle
```diff
@@ -55,14 +55,21 @@ jenkinsPlugin {
 sourceCompatibility = '1.7'
 
 dependencies {
-  compile 'org.jenkins-ci.lib:dry-run-lib:0.1'
-  compileOnly 'org.jenkins-ci:symbol-annotation:1.3'
-  jenkinsPlugins 'org.jenkins-ci.plugins:structs:1.3@jar'
+  jenkinsPlugins 'org.jenkins-ci.plugins:dry-run:0.9'
```
wolfs: The new version of the `dry-run` plugin seems to depend on the `maven-plugin`. Can we exclude the `maven-plugin` as a dependency? Something like
```
configurations {
  all {
    exclude group: 'org.jenkins-ci.main', module: 'maven-plugin'
  }
}
```

wolfs:
Maybe an approach as shown here would be better: https://jenkins.io/blog/2016/04/21/dsl-plugins/

wolfs:
There is also https://github.com/jenkinsci/pipeline-maven-plugin, which seems to use the `Step` API.

wolfs:
I'll be closing the PR, given that there hasn't been any recent activity. Is the PR still interesting? If yes, please open a new one based on current `main`. Thank you!

PR #63, opened 2018-12-29
[JENKINS-53575] Configuration-as-Code compatibility

darxriggs:
This fixes problems with JCasC for configuration import & export.

Also see the respective issue in JCasC - jenkinsci/configuration-as-code-plugin/issues/460.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,8 @@ plugins {
   id 'codenarc'
 }
 
+apply from: "$rootDir/gradle/jcascTest.gradle"
+
 group = 'org.jenkins-ci.plugins'
 description = 'This plugin adds Gradle support to Jenkins'
 
@@ -18,7 +20,7 @@ if (ciBuild) {
 
 jenkinsPlugin {
   // Version of Jenkins core this plugin depends on.
-  coreVersion = '2.60.3'
+  coreVersion = '1.642.1'
 
   // Human-readable name of plugin.
   displayName = 'Gradle Plugin'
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -18,7 +18,7 @@ if (ciBuild) {
 
 jenkinsPlugin {
   // Version of Jenkins core this plugin depends on.
-  coreVersion = '1.642.1'
+  coreVersion = '2.60.3'
```
wolfs: Can we get around rising the base version by adding a new test task which tests against a different Jenkins version? Can you try to make this happen?

I nearly prefer having a unit test which tests for the specific problem and not incrementing the base verison.

src/main/java/hudson/plugins/gradle/GradleInstallation.java
```diff
@@ -111,17 +108,16 @@ public String getDisplayName() {
             return Collections.singletonList(new GradleInstaller(null));
         }
 
-        // for compatibility reasons, the persistence is done by GradleBuilder.DescriptorImpl
+        // for compatibility reasons, the persistence is done by Gradle.DescriptorImpl
 
         @Override
         public GradleInstallation[] getInstallations() {
-            return gradleDescriptor.getInstallations();
+            return Jenkins.getActiveInstance().getDescriptorByType(Gradle.DescriptorImpl.class).getInstallations();
```
wolfs: Can we keep the injected field and use the static lookup only if the field is `null`? This logic should probably also extracted into a `getGradleDescriptor` method.

build.gradle
darxriggs: I added a "jcascTest" configuration + source set + task to make it happen. The commit comment describes all the details.

I am not 100% satisfied. The IDEA integration cannot handle this setup properly. But that's the best I can come up with, given my current Gradle knowledge and available time.

PR #64, opened 2019-01-05
Improve test source set handling

wolfs:

PR #65, opened 2019-01-06
Fix server task

darxriggs:
Running `gradle server` failed with the following error before this change:
> Failed Loading plugin gradle
> java.io.IOException: Dependency structs (1.3) doesn't exist

Transitive plugin dependency resolution has been fixed in org.jenkins-ci.jpi 0.23.1.
Excerpt from CHANGELOG: "plugin dependencies must not longer be specified with artifact only notation"

This jpi plugin fix can now be used to declare the dependencies without the artifact only notation. As a result the structs plugin is also provided correctly as plugin for the server task.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,13 +58,12 @@ sourceCompatibility = '1.7'
 
 dependencies {
   compile 'org.jenkins-ci.lib:dry-run-lib:0.1'
-  compileOnly 'org.jenkins-ci:symbol-annotation:1.3'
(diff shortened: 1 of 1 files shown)


wolfs:
I did the changes you are suggesting on `master` (https://github.com/jenkinsci/gradle-plugin/pull/64), can you check if the `server` tasks works now?

wolfs:
Thanks for spotting this, looks like I missed it. I'll fix it.

darxriggs:
@wolfs as you incorporated the changes in this pull request yourself via other commits this pull request can be closed unmerged.

PR #66, opened 2019-02-21
Support detecting build scans from non-gradle build steps

mark-vieira:
This PR adds support for detecting build scan URLs in build logs
from jobs that don't use the Gradle builder. This is often the case
for projects that use some kind of additional wrapper around Gradle and
therefore rely on the Shell builder for executing builds.

This PR also simplifies support for jobs that publish multiple scans
by creating a single Action and consolidating all URLs under a single
job summary section.


build.gradle
```diff
@@ -22,7 +22,7 @@ if (ciBuild) {
 
 jenkinsPlugin {
   // Version of Jenkins core this plugin depends on.
-  coreVersion = '1.642.1'
+  coreVersion = '1.651.3'
```
wolfs: Is the core update required?

src/main/java/hudson/plugins/gradle/BuildScanAction.java
```diff
@@ -2,6 +2,12 @@
 
 import hudson.model.Action;
 
+/**
+ * @deprecated This class remains only for backwards compatibility so that actions persisted for old builds will
+ * continue to render correctly. At some point this should probably be removed.
+ *
+ * @see BuildScansAction
```
wolfs: Can you do the data migration of this class to the new class? Some hints here: https://wiki.jenkins.io/display/JENKINS/Hint+on+retaining+backward+compatibility

src/main/java/hudson/plugins/gradle/GradleConsoleAnnotator.java
```diff
@@ -33,15 +37,19 @@ protected void eol(byte[] b, int len) throws IOException {
         // trim off CR/LF from the end
         line = trimEOL(line);
 
-        if (line.startsWith(":") || line.startsWith("> Task :"))
-            // put the annotation
-            new GradleTaskNote().encodeTo(out);
+        if (usesGradleBuilder) {
```
wolfs: I think we should add the console decoration if when not using the Gradle builder. Any reason this should not be done?

src/main/java/hudson/plugins/gradle/GradleConsoleLogFilter.java
```diff
@@ -0,0 +1,49 @@
...
+import hudson.console.ConsoleLogFilter;
+import hudson.model.FreeStyleBuild;
+import hudson.model.Run;
+import hudson.tasks.Builder;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+@Extension
+public class GradleConsoleLogFilter extends ConsoleLogFilter implements BuildScanPublishedListener {
```
wolfs: IIUC, this applies the filter to all console output of all builds on the Jenkins instance? I would prefer being able to use this as an opt-in, maybe via a build wrapper. There are probably some performance implications for parsing all logs generating on Jenkins.

src/test/groovy/hudson/plugins/gradle/BuildScanIntegrationTest.groovy
```diff
@@ -1,43 +1,72 @@
...
-        "1.6" | "3.4" | "-Dscan"
-        "1.8" | "4.0" | "--scan"
+        "1.6"            | "3.4"         | "-Dscan"
+        "1.8"            | "4.0"         | "--scan"
+    }
+
+    def 'build scan is discovered when using non-gradle build step'() {
+        given:
+        FreeStyleProject p = j.createFreeStyleProject()
+        p.setScm(new ExtractResourceSCM(this.class.getResource('/gradle/wrapper.zip')))
```
wolfs: I would prefer not using a checked in zip file as a resource. We could pack up the checked in wrapper in the build and provide it as a resource or install Gradle via the existing tool support used in other tests.

src/main/java/hudson/plugins/gradle/GradleConsoleAnnotator.java
mark-vieira: Mainly because it seemed likely that arbitrary builds might contain lines prefixed with `:` that aren't necessarily Gradle tasks. It seemed a common enough token where folks might start to get funky highlighting in existing builds.

build.gradle
mark-vieira: It was to install the `Git` plugin which I used for local testing. It's the current 1.6xx LTS so seemed more appropriate anyhow vs 1.642.1 which is a bit of an arbitrary version. I don't think from an API standpoint I need anything from that newer version though. We could back this out.

src/main/java/hudson/plugins/gradle/BuildScanAction.java
mark-vieira: I did some refactoring here, but now I'm not sure there's a simple solution. The problem here is that the arity has changed. The old implementation supported multiple actions per build, but the new does not. I'm not sure there's a simple way to take n `BuildScanAction` and turn them into a single `BuildScansAction`. As it is now, my naive refactor will create multiple actions again for old builds which might result in a poor UI experience.

src/main/java/hudson/plugins/gradle/GradleConsoleLogFilter.java
mark-vieira: I've reimplemented this as a `BuildWrapper`.

src/test/groovy/hudson/plugins/gradle/BuildScanIntegrationTest.groovy
mark-vieira: Good idea. The build now packages up the existing wrapper and includes in on the test classpath.

src/main/java/hudson/plugins/gradle/BuildScanAction.java
wolfs: There is the possibility to use an `Initializer`: https://github.com/jenkinsci/run-selector-plugin/blob/854deb121eace1d0f4a90664a6d2d2ad733e0d61/src/main/java/hudson/plugins/copyartifact/CopyArtifact.java#L148-L172

Please also add some test for the migration in `CompatibilityTest`.

src/main/java/hudson/plugins/gradle/GradleConsoleAnnotator.java
wolfs: Fine with me. The plugin should probably start using an init script for getting the information as Teamcity does. That probably won't help for arbitrary build step logs I guess...

src/main/java/hudson/plugins/gradle/Gradle.java
```diff
@@ -352,6 +349,7 @@ private boolean performTask(boolean dryRun, AbstractBuild<?, ?> build, Launcher
         }
     }
 
+
```
wolfs: Nit: double newline.

src/main/java/hudson/plugins/gradle/BuildScanBuildWrapper.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public DescriptorImpl getDescriptor() {
+        return (DescriptorImpl) super.getDescriptor();
+    }
+
+    @Extension
+    public static final class DescriptorImpl extends BuildWrapperDescriptor {
+
+        public DescriptorImpl() {
```
wolfs: No need to add a constructor here. The class should be inferred from the outer class.

src/main/java/hudson/plugins/gradle/BuildScanBuildWrapper.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @CheckForNull
+    @Override
+    public ConsoleLogFilter createLoggerDecorator(@Nonnull Run<?, ?> build) {
+        return new GradleConsoleLogFilter();
+    }
+
+    @Override
+    public DescriptorImpl getDescriptor() {
```
wolfs: No need to overwrite `getDescriptor()`.

mark-vieira: Ok, I wasn't sure if this was necessary. Is there a reason this is done so in `Gradle`?

wolfs: If the descriptor has persisted data itself, it needs to use `load` and `save`. This is true for `Gradle`, though the descriptor here has no data.

That is what I figured from the Javadoc of `Descriptor`:
```
 * <h2>Persistence</h2>
 * <p>
 * {@link Descriptor} can persist data just by storing them in fields.
 * However, it is the responsibility of the derived type to properly
 * invoke {@link #save()} and {@link #load()}.
```

wolfs: Ah sorry, my reply was for why `Gradle` overrides the constructor. I don't know why it overrides the getter...

src/main/java/hudson/plugins/gradle/BuildScanAction.java
mark-vieira: I'm concerned how costly it might be to go back and migrate this for all build. We aren't just migrating project config, these are actions attached to _actual_ build invocations. Going back and updating the entire Jenkins history might not be worth the effort here.

wolfs: OK, I am fine with leaving it the way it is now - i.e. the name of the action does not change, we only store a list instead of a single element. We could change the name as well (via `Items.XSTREAM2.addCompatibilityAlias`), though I don't think it is necessary.

We still need a test. I think currently it won't work since the list will be null when loaded from an old version of `BuildScanAction`.

mark-vieira: Yeah, as I mentioned, I had initially done that and implemented `readResolve()` to populate the list property if `scanUrl` was non-null but that breaks down because of the arity issue. We would create multiple of these actions which would then render oddly, as the updated UI assumes 0..n build scans.

I've simply reverted back to the initial compatibility method which just keeps the old implementation around. So for old builds, we'll use the old action and UI, and for new builds we'll use the new one. I simply don't have a better idea here that doesn't include a costly migration.

src/main/java/hudson/plugins/gradle/Gradle.java
mark-vieira: Fixed.

src/main/java/hudson/plugins/gradle/BuildScanAction.java
mark-vieira: Implementation is per our discussion. I've also added test coverage for compatibility.

PR #69, opened 2019-04-10
[WIP] Java 11 readiness: build both on JDK8 and JDK11

batmat:
Work in progress to make sure this plugin is constantly checked both building with JDK8 and JDK11.

This goes with the Jenkins Java 11 General Availability announcement made back in March, and we're making a pass to check plugins are looking good on a JDK11.

([to understand the Jenkinsfile content](https://wiki.jenkins.io/display/JENKINS/Java+11+Developer+Guidelines#Java11DeveloperGuidelines-MakesureyourpluginistestedinContinuousIntegrationonJava8andJava11atthesametime))

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin()
+buildPlugin(configurations: buildPlugin.recommendedConfigurations())


wolfs:
I just looked at the source code of the configuration. IIUC, the `jenkinsVersion` only has an effect on Maven builds. The `gradle-plugin` is built with Gradle. Does that mean that the change has no effect apart from doing the same thing more often? Or am I missing something?

batmat:
Yes, I do not think this is applicable to gradle builds. Sorry I assumed this should work too given there was a Jenkinsfile already but didn't actually check if the `buildPlugin` step handles gradle or not.

PR #75, opened 2019-07-19
Support variables in build file location

pavelkokush:
Resolve variables for **buildFile** in **getPossibleWrapperLocations** method (same as in **Gradle.perform** method)
Now it is not possible to calculate gradle wrapper location if **buildFile** contains env variables.


wolfs:
@pavelkokush Thank you for your pull request!

PR #76, opened 2019-08-15
WIP - Demo build for buildPluginWithGradle()

oleg-nenashev:
This is a reference implementation and a Demo makefile for https://github.com/jenkins-infra/pipeline-library/pull/104 . The makefile uses Jenkinsfile Runner, but it works in a similar way as ci.jenkins.io in this case


Makefile
```diff
@@ -0,0 +1,11 @@
+clean:
```
wolfs: What is the makefile used for? Will this be called somehow on CI?

oleg-nenashev:
@wolfs Once the upstream pull request is ready, I will move it to the Pipeline Library as an integration test. I do not expect the Makefile code to remain in the final version of the PR

oleg-nenashev:
The new step has been integrated in https://github.com/jenkins-infra/pipeline-library/releases/tag/1.4.0 . I have retriggered the build and removed the Makefile. It will be later added as integration test to the library directly

PR #77, opened 2019-09-02
Return build scans list when calling findBuildScans()

slowr:
Currently there is no way to retrieve the build scan URLs on post actions in order to send them to a chat application.

Before merging would love to know how to add a `BuildScanIntegrationTest` to verify my changes.

### DIFF
--- a/src/main/java/hudson/plugins/gradle/BuildScanPublisher.java
+++ b/src/main/java/hudson/plugins/gradle/BuildScanPublisher.java
@@ -12,6 +12,7 @@ import org.kohsuke.stapler.DataBoundConstructor;
 
 import javax.annotation.Nonnull;
 import java.io.BufferedReader;
+import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/gradle/BuildScanPublisher.java
```diff
@@ -42,7 +43,8 @@ protected Void run() throws Exception {
             ) {
                 lines.forEach(scanner::scanLine);
             }
-            return null;
+
+            return run.getAction(BuildScanAction.class).getScanUrls();
```
wolfs: This may give a null pointer exception if no scans are found. How about you use a composite listener which keeps track of the found build scans and is passed into `BuildScanLogScanner`.

It would also be fine if `DefaultBuildScanPublishedListener` can be queried for the found build scans.

Finally, a null check is also good enough for now.

slowr: I chose to go with the null check as I don't have an environment right now to run the integration tests and it seemed the most straight forward. please take a look on the new changes.

PR #80, opened 2019-09-05
README: Add badges for the plugin version and number of installations

oleg-nenashev:


wolfs:
Thank you @oleg-nenashev!

PR #83, opened 2019-09-17
Reference the Release Drafter docs from GitHub Actions

oleg-nenashev:
GitHub Actions are enabled now, but they did not pick up the workflow definition automatically.
Looks like we need a commit to trigger the workflow creation.

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -1,3 +1,5 @@
+# Automates creation of Release Drafts using Release Drafter
+# More Info: https://github.com/jenkinsci/.github/blob/master/.github/release-drafter.adoc
 name: Release Management
 
 on:


oleg-nenashev:
```
08:55:09.968Z ERROR event: Cannot read property 'login' of null (id=960355aa-000c-40cc-8022-c170ce26de0f)
  TypeError: Cannot read property 'login' of null
      at acc.push.categorizedPullRequests.(anonymous function).map.pullRequest (/app/lib/releases.js:153:43)
      at Array.map (<anonymous>)
      at orderedCategories.reduce (/app/lib/releases.js:149:12)
      at Array.reduce (<anonymous>)
      at generateChangeLog (/app/lib/releases.js:136:6)
      at module.exports.generateReleaseInfo (/app/lib/releases.js:179:17)
      at app.on (/app/index.js:69:25)
  --
  event: {
    "event": "push",
    "id": "960355aa-000c-40cc-8022-c170ce26de0f",
    "repository": "jenkinsci/gradle-plugin"
  }
```

So we have an answer while it fails

oleg-nenashev:
```
$AUTHOR: pullRequest.author.login
```

So there is an PR author without a login, hmm. Deleted user maybe?

PR #91, opened 2019-12-21
Consistent use of single quotes

darxriggs:


wolfs:
Thank you @darxriggs!

PR #93, opened 2019-12-21
Upgrade Gradle to 6.0.1

darxriggs:
The build-scan plugin id has changed and the plugin has to be configured differently starting with Gradle 6, see https://docs.gradle.com/enterprise/gradle-plugin/#upgrading_to_gradle_6

The FindBugs plugin was removed from Gradle 6, see https://github.com/gradle/gradle/issues/6652
Therefore replacing it with SpotBugs, see https://plugins.gradle.org/plugin/com.github.spotbugs


settings.gradle
```diff
@@ -1,3 +1,7 @@
+plugins {
+    id 'com.gradle.enterprise' version '3.1.1'
```
wolfs: Could you move the configuration of the build scan plugin also to `settings.gradle`? I think the other usage of `ciBuild` (`ignoreFailures = ciBuild`) can be removed from `build.gradle`.

darxriggs: I think `ignoreFailures = ciBuild` should be kept, otherwise on failure the test results would not be recorded - see [buildPluginWithGradle](https://github.com/jenkins-infra/pipeline-library/blob/1.4.1/vars/buildPluginWithGradle.groovy#L66). There is no central configuration provided by the infrastructure library as it is available for Maven. We already had a discussion at https://github.com/jenkins-infra/pipeline-library/pull/88#issuecomment-521341959.

wolfs: I was hoping that the test results are captured even if one of the builds fails... `ignoreFailures = false` is bad, because the build cache would then cache the result of a test, even if there are test failures (since the test task succeeded). That is why I would prefer the test task to fail if there are test failures.

Since we are now not using the build cache on CI, it is safe to keep the `ignoreFailures = false` flag.

wolfs:
Thank you, @darxriggs!

PR #138, opened 2022-01-08
Find build scans in console logs

wolfs:
So the links to the build scans show up on the build page.


wolfs:
This doesn't work like this, let's close the PR right now.

PR #150, opened 2022-06-03
Allow Build Scan publishing without changing project configurations

alextu:
Add the ability to enable Build Scans for both Gradle and Maven builds without having to change any pipeline. The feature is enabled by setting specific environment variables, see README.adoc for more information.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -15,6 +15,8 @@ http://www.gradle.org/[Gradle] is managed as another tool inside Jenkins (the sa
 
 It also allows detecting https://gradle.com/build-scans/[Build Scans] in arbitrary console logs, for Maven and Gradle builds and display them in the Jenkins UI.
 
+The plugin also allows for injecting Build Scans into both Maven and Gradle builds by setting some environment variables, see link:#_injecting_build_scans[Injecting Build Scans].
+
 In order to release this plugin have a look at link:RELEASING.md[here].
 
 == Configuration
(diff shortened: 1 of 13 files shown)


src/main/resources/scripts/init-script.gradle
```diff
@@ -0,0 +1,175 @@
+import org.gradle.util.GradleVersion
```
britter, line 1: This should be in "our" package so it does not clash with something else. Move this to `src/main/resources/hudson/plugins/gradle/injection"

src/main/java/hudson/plugins/gradle/injection/SetMavenOpts.java
```diff
@@ -0,0 +1,30 @@
+package hudson.plugins.gradle.injection;
+
+import hudson.EnvVars;
+import jenkins.security.MasterToSlaveCallable;
+
+import java.util.List;
+import java.util.logging.Logger;
+
+public class SetMavenOpts extends MasterToSlaveCallable<Void, RuntimeException> {
```
britter, line 9: We should try whether we can set the environment from `MavenBuildScanInjection` by calling 
```java
node.getNodeProperties().add(new EnvironmentVariablesNodeProperty(new EnvironmentVariablesNodeProperty.Entry("MAVEN_OPTS", "the value")));
```

src/test/groovy/hudson/plugins/gradle/JavaGitContainer.java
```diff
@@ -0,0 +1,13 @@
+package hudson.plugins.gradle;
+
+import org.jenkinsci.test.acceptance.docker.DockerFixture;
+import org.jenkinsci.test.acceptance.docker.fixtures.JavaContainer;
+
+/**
+ * Fixture capable of running java programs and git over ssh.
+ *
```
britter, line 8:
```suggestion
 * Fixture capable of running java programs and git over ssh.
 *
 * Copied from pipeline-maven-plugin.
 * 
```

README.adoc
```diff
@@ -86,6 +88,49 @@ The `withGradle` wrapper should be used instead, since it also deals well with p
...
+|===
+
+In addition to the variables above you might want to set the `GRADLE_ENTERPRISE_ACCESS_KEY` variable if you're Gradle Enterprise server requires authentication for publishing Build Scans. Refer to the https://docs.gradle.com/enterprise/gradle-plugin/#via_environment_variable[Gradle Enterprise Gradle plugin manual] and the https://docs.gradle.com/enterprise/maven-extension/#via_environment_variable[Gradle Enterprise Maven Extension manual] for more information.
+
+=== Supported Job types
+
+This feature works for the following Jenkins jobs types:
+
+-
+
```
britter, line 128:
```suggestion
```

britter (approved):

PR #152, opened 2022-06-06
Fix build scan injection maven itest (plus other minor changes)

alextu:
Prod code:
- Attempt to get existing `MAVEN_OPTS` on agent and set with `EnvironmentVariablesNodeProperty`, works on test, did not get a chance to test on real setup
- Move init script to `/hudson/plugins/gradle/injection/init-script.gradle`

Regarding tests adjustments:
- Copy mvn ext libs on `processResources`
- Replace docker container agent in itest by simple jar agent and remove unnecessary dependencies
- Avoid using spaces in temp directories, not supported when setting `MAVEN_OPTS`


src/main/java/hudson/plugins/gradle/injection/MavenBuildScanInjection.java
```diff
@@ -63,12 +63,22 @@ private void injectMavenExtension(FilePath rootPath) {
...
             throw new IllegalStateException(e);
         }
     }
 
+    void appendMavenOpts(Node node, List<String> mavenOptsKeyValuePairs) throws IOException, InterruptedException {
+        EnvVars nodeEnvVars = EnvVars.getRemote(node.getChannel());
+        String mavenOpts = String.join(" ", mavenOptsKeyValuePairs);
+        String mavenOptsEnvName = "MAVEN_OPTS";
+        if (nodeEnvVars.get(mavenOptsEnvName) != null) {
+            mavenOpts = nodeEnvVars.get(mavenOptsEnvName) + " " + mavenOpts;
```
britter, line 77: We need to check here whether we've not already appended (e.g. because of another configuration change that happened before)

src/main/java/hudson/plugins/gradle/injection/MavenBuildScanInjection.java
```diff
@@ -63,29 +73,44 @@ private void injectMavenExtension(FilePath rootPath) {
...
             deleteResourceFromAgent(GE_MVN_LIB_NAME, rootPath);
             deleteResourceFromAgent(CCUD_LIB_NAME, rootPath);
-            rootPath.act(new ClearMavenOpts(
-                    MAVEN_EXT_CLASS_PATH_PROPERTY_KEY,
-                    GRADLE_SCAN_UPLOAD_IN_BACKGROUND_PROPERTY_KEY,
-                    GRADLE_ENTERPRISE_ALLOW_UNTRUSTED_SERVER_PROPERTY_KEY,
-                    GRADLE_ENTERPRISE_URL_PROPERTY_KEY
-            ));
+            String mavenOpts = clearMavenOpts.removeSystemProperties(getMavenOpts(node));
+            setMavenOpts(node, mavenOpts);
```
britter, line 92: I think this object could do more. Getting and setting should also be part of it. I think it should look like this:

```java
class MavenOptsSetter {

    MavenOptsSetter(String... systemPropertyKeys) {
        //...
    }

    void appendIfMissing(Node node, String mavenOptsKeyValuePairs) {
        // get MAVEN_OPTS from node, check if we need to append and if so append them
    }

    void remove(Node node) {
        // get MAVEN_OPTS from the node, remove our keys and send the result back to the node
    }
}
```

britter: With this you could even write a unit test that makes sure we don't append multiple times and we don't destroy pre-existing values of `MAVEN_OPTS`.

britter (approved):
LGTM but I would move more code out of the injection class into a dedicated class. See my comment.

PR #156, opened 2022-06-18
Allow opt-out of build scan injection

sghill:
Hi,

Would you be open to opting out of build scan injection? I'm interested in this for two reasons:
1. We run a custom distribution that already has build scans configured.
2. We don't pass `$HOME` to our agents, so we started getting `Error: HOME is not set` on every agent connection.

To avoid the warning, I have started setting `JENKINSGRADLEPLUGIN_BUILD_SCAN_OVERRIDE_HOME=${mktemp -d)`, but I would prefer to set a property to opt-out.

The approach here would allow opting-out by setting a system property, similar to other features in Jenkins:

```
-Dhudson.plugins.gradle.injection.BuildScanInjectionListener.disabled=true
```

This isn't particularly testable, so I'm happy to use a different implementation if preferred. We could instead introduce a Guice module that injects a system property.

- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/java/hudson/plugins/gradle/injection/BuildScanInjectionListener.java
+++ b/src/main/java/hudson/plugins/gradle/injection/BuildScanInjectionListener.java
@@ -17,6 +17,7 @@ import java.util.logging.Logger;
 public class BuildScanInjectionListener extends ComputerListener {
 
     private static final Logger LOGGER = Logger.getLogger(BuildScanInjectionListener.class.getName());
+    private static final boolean DISABLED = Boolean.getBoolean(BuildScanInjectionListener.class.getName() + ".disabled");
 
     private final List<BuildScanInjection> injections = Arrays.asList(
             new GradleBuildScanInjection(),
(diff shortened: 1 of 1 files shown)


sghill:
@wolfs would you be open to an opt-out flag for build scan injection?

sghill:
Thanks for the quick response @wolfs. It doesn't fail the connection, but it does log at the warn level each time. This gets fairly noisy in large deployments.

I'd like to avoid the warning without suppressing the log category or passing home/the override directory to the agent.

sghill:
@wolfs #157 should work.

It seems conceptually simpler to me to allow an opt-out of some kind (setting the `injections` list to empty?) rather than always making the attempt to resolve directories, copy files, and conditionally log. Ultimately achieves my goal though, so happy to run with it if that's what you prefer.

wolfs:
@sghill #157 has been released as part of [1.39.2](https://github.com/jenkinsci/gradle-plugin/releases/tag/gradle-1.39.2).

sghill:
Thanks @wolfs!

PR #161, opened 2022-07-05
Get available embedded GE/CCUD jars

alextu:
This is a first step to avoid the hardcoded version.
Next step would be to download the specified version but that requires more thinking.


src/main/java/hudson/plugins/gradle/injection/CpPatternResource.java
```diff
@@ -0,0 +1,72 @@
...
+            }
+        } catch (URISyntaxException | IOException e) {
+            throw new IllegalStateException(notFound, e);
+        }
+        throw new IllegalStateException(notFound);
+    }
+
+    private Optional<String> findInJar(String directoryName, URL url) throws IOException {
+        String path = url.getPath();
+        String jarPath = path.substring(5, path.indexOf("!"));
```
jprinet, line 59: My understanding is that we want the jar file name without protocol and internal resource.
I would extract this to a method with a meaningful name to clarify the intent?

jprinet (approved):

src/main/java/hudson/plugins/gradle/injection/CpPatternResource.java
alextu: yeah it basically transforms from `file:/path/to/some.jar!some/dir` to `/path/to/some.jar`

alextu: I've extracted a method and created a simple test for doc

PR #210, opened 2022-11-02
Support TeamCity CI integration

guylabs:


welandaz (approved):

settings.gradle
```diff
@@ -12,26 +13,30 @@ if (!JavaVersion.current().java11) {
...
-      ifNonNull(System.getenv('GIT_COMMIT')) { value('git-commit-id', it) }
-      ifNonNull(System.getenv('BUILD_URL')) { link('Jenkins build', it) }
-
-      termsOfServiceUrl = 'https://gradle.com/terms-of-service'
-      termsOfServiceAgree = 'yes'
-      publishAlways()
-    } else {
-      tag 'LOCAL'
+    capture { taskInputFiles = true }
+    uploadInBackground = !gradle.ciBuild
```
facewindu, line 21: shouldn't this take `ciTeamCityBuild` into account too ?

settings.gradle
```diff
@@ -12,26 +13,30 @@ if (!JavaVersion.current().java11) {
 }
 
 gradle.ext.ciBuild = System.getenv()['JENKINS_URL'] ? true : false
```
facewindu, line 15: should this now be better named with Jenkins in it ?

settings.gradle
guylabs: Done

settings.gradle
guylabs: Renamed

build.gradle
```diff
@@ -158,10 +158,10 @@ test {
   systemProperties['hudson.model.DownloadService.noSignatureCheck'] = 'true'
   // override default timeout
   systemProperties['jenkins.test.timeout'] = 300
-  ignoreFailures = gradle.ciBuild
+  ignoreFailures = gradle.isCi
```
wolfs, line 161: I think we only want to ignore test failures on Jenkins and not on Teamcity.

README.adoc
```diff
@@ -6,7 +6,6 @@
 image:https://img.shields.io/jenkins/plugin/v/gradle.svg[Jenkins Plugin,link=https://plugins.jenkins.io/gradle]
 image:https://img.shields.io/github/release/jenkinsci/gradle-plugin.svg?label=changelog[link="https://github.com/jenkinsci/gradle-plugin/releases/latest"]
 image:https://img.shields.io/jenkins/plugin/i/gradle.svg?color=blue[Jenkins Plugin Installs,link=https://plugins.jenkins.io/gradle]
-image:https://ci.jenkins.io/buildStatus/icon?job=Plugins/gradle-plugin/master[Build Status,link=https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgradle-plugin/branches/]
```
wolfs, line 9: Why remove the link to the Jenkins build here?

facewindu (approved):

README.adoc
guylabs: Because it has been red most of the time in the past due to the flaky browser tests. We can keep it but having it red sends out a wrong signal and therefore I'd rather remove it.

build.gradle
guylabs: Thinking about this some more, shouldn't we just remove that completely and have it fail always? Why has this been added in the first place? To make the Jenkins build green even though there were test failures ?

wolfs: Jenkins builds have three possible outcomes: red, yellow and green/blue. The yellow state is what GE calls a verification failure. In order for Jenkins to show a verification failure for Gradle builds, the build needs to succeed and there need to be failing tests. That is why I ignored failures when running on CI.

README.adoc
wolfs: Alright, if you prefer not having the link at all that is fine with me.

build.gradle
guylabs: Understood. Changed it.

wolfs (approved):
LGTM!

README.adoc
c00ler: @guylabs I think we need to keep it and just move browser tests to TeamCity and don't run them on Jenkins. Tests are flaky due to the infra, on TeamCity the tests should be stable because we have control over the infrastructure.

settings.gradle
```diff
@@ -11,27 +12,32 @@ if (!JavaVersion.current().java11) {
   throw new GradleException('Build requires Java 11')
 }
 
-gradle.ext.ciBuild = System.getenv()['JENKINS_URL'] ? true : false
+gradle.ext.ciJenkinsBuild = System.getenv()['JENKINS_URL'] ? true : false
```
c00ler, line 15:
```suggestion
gradle.ext.ciJenkinsBuild = System.getenv('JENKINS_URL') ? true : false
```

c00ler (approved):

sorin-florea (approved):

README.adoc
lemeurherve: @c00ler what kind of issues have you encountered on Jenkins infrastructure? Is there something we could do?

c00ler: Hi @lemeurherve, thank you for looking into it! 
We are running many browser tests, with different versions of Jenkins, and were facing some performance constraints, due to the agent machines' size. But this is something definitely expected. 
Another issue that we've seen quite frequently are failures due to the browser driver, for example [here](https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgradle-plugin/detail/master/181/tests), or [here](https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgradle-plugin/detail/PR-210/2/tests/). It's hard to say at the moment what's the cause of these issues and if anything can be done to improve the stability.

lemeurherve: @c00ler thanks for the feedback :)

About the agent size, you can use the beefier [highmem](https://github.com/jenkins-infra/jenkins-infra/blob/93f7c1536b726d95cc9e6dec78779e56bad323c0/hieradata/clients/azure.ci.jenkins.io.yaml#L322) node label for your browser tests.

For the failures you encountered, would you mind [opening an help desk issue](https://github.com/jenkins-infra/helpdesk) so we can take a more closer look at them please?

lemeurherve: I realized you can't choose the label with buildPluginWithGradle, I'll implement it so you can set an `highmem` label.

PR #185, opened 2022-09-27
Form based injection configuration

c00ler:
## Description

This pull request changes the Build Scans injection configuration source from environment variables to the form in the global `Configure System` menu. It works the same way as it was working with the environment variables. There is one global toggle, that enables the Build Scans injection feature. 

Screenshot 2022-10-03 at 23 23 00

Gradle Enterprise server URL is mandatory. It applies to both Gradle and Maven injections. 

Gradle Build Scans injection is enabled by setting the version of the Gradle Enterprise Gradle Plugin

Screenshot 2022-10-03 at 23 25 01

Maven Build Scans injection is enabled by setting the version of the Gradle Enterprise Maven Extension

Screenshot 2022-10-03 at 23 25 13

## TODO

This pull request is missing updates to the documentation. They have to be done after the following PR https://github.com/jenkinsci/gradle-plugin/pull/186 is merged.

## Checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #211, opened 2022-11-03
Revert jpi ext change

timja:
Hi 

I'm not sure why this change was made other than it was in commit [`07a9c1d` (#185)](https://github.com/jenkinsci/gradle-plugin/pull/185/commits/07a9c1df5975955b5c39dc74adb190839d41b034)

This breaks:
* Installing the Gradle plugin via the Acceptance Test Harness, see https://github.com/jenkinsci/acceptance-test-harness/issues/985
* Adding this plugin as a dependency to another plugin and running it with `mvn hpi:run`

It would be appreciated if a quick release could be done so we can have the acceptance tests passing again, if it was just the gradle plugin failing we would skip the tests but it also prevents any plugins working that depend on the gradle plugin...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,6 @@ jenkinsPlugin {
 
   // Plugin ID, defaults to the project name without trailing '-plugin'
   shortName = 'gradle'
-  
   fileExtension = 'hpi'
 
   compatibleSinceVersion = '1.0'
--- a/build.gradle
+++ b/build.gradle
@@ -36,8 +36,6 @@ jenkinsPlugin {
   // Plugin ID, defaults to the project name without trailing '-plugin'
   shortName = 'gradle'
 
-  fileExtension = 'jpi'
-
   compatibleSinceVersion = '1.0'
 
   developers {
--- a/build.gradle
+++ b/build.gradle
@@ -35,6 +35,7 @@ jenkinsPlugin {
 
   // Plugin ID, defaults to the project name without trailing '-plugin'
   shortName = 'gradle'
+
   fileExtension = 'hpi'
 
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -36,8 +36,6 @@ jenkinsPlugin {
   // Plugin ID, defaults to the project name without trailing '-plugin'
   shortName = 'gradle'
 
-  fileExtension = 'jpi'
```
c00ler, line 39: @timja Sorry for introducing this breaking change. Yes, we can revert the extension back to `.hpi` and then release a new version with this change. 

Maybe it's better to specify `fileExtension = 'hpi'` explicitly? To be sure that we don't depend on any implicit behavior of the plugin. WDYT?

build.gradle
```diff
@@ -35,8 +35,8 @@ jenkinsPlugin {
 
   // Plugin ID, defaults to the project name without trailing '-plugin'
   shortName = 'gradle'
-
-  fileExtension = 'jpi'
+
```
timja, line 38:
```suggestion

```

build.gradle
```diff
@@ -35,8 +35,7 @@ jenkinsPlugin {
 
   // Plugin ID, defaults to the project name without trailing '-plugin'
   shortName = 'gradle'
-
-  fileExtension = 'jpi'
+  fileExtension = 'hpi'
```
timja, line 38:
```suggestion

  fileExtension = 'hpi'
```

build.gradle
timja: sure done

c00ler: Thank you! I will release a new version with this change today.

PR #214, opened 2022-11-09
Skip auto-injection if older version of maven-plugin is used

c00ler:
This PR downgrades the optional `maven-plugin` dependency version. The case when this functionality may be needed is the following. Users already use an older version of the `maven-plugin` and want to add Gradle builder to their Jenkins instance. They are not interested in the Build Scans auto-injection functionality. Without this change, users would be forced to update the `maven-plugin`. With this change, users can don't need to update it. The `maven-plugin` integration will be automatically disabled if an unsupported `maven-plugin` version is detected.


src/main/java/hudson/plugins/gradle/injection/maven/GradleEnterpriseExtensionsContributorFactory.java
```diff
@@ -44,6 +49,15 @@ public PlexusModuleContributor createFor(AbstractBuild<?, ?> build) {
                 return EMPTY_CONTRIBUTOR;
             }
 
+            if (!isSupportedMavenPluginVersion()) {
+                LOGGER.log(
+                    Level.WARNING,
+                    "An older maven-plugin version is used. Please use at least the version {0} for auto-injection to work",
```
guylabs, line 55: Can we list the current Maven plugin version ?

I'd then also write it that way:

```
Detected Maven Integration plugin version x.x.x. For auto-injection of the Gradle Enterprise Maven extensions, version x.x.x is required. Please upgrade the version of the Maven Integration plugin.
```

guylabs (changes requested):
Small change. Nice that you were able to do the test setup with the version override!

sorin-florea (approved):
LGTM, nice work with the tests.

welandaz (approved):

src/main/java/hudson/plugins/gradle/injection/maven/GradleEnterpriseExtensionsContributorFactory.java
c00ler: Done

guylabs (approved):

PR #215, opened 2022-11-10
Improve build scan link capturing

sorin-florea:

Fixes https://issues.jenkins.io/browse/JENKINS-69934

### DIFF
--- a/src/main/java/hudson/plugins/gradle/BuildScanLogScanner.java
+++ b/src/main/java/hudson/plugins/gradle/BuildScanLogScanner.java
@@ -1,31 +1,37 @@
 package hudson.plugins.gradle;
 
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class BuildScanLogScanner {
+    private static final Pattern BUILD_SCAN_PATTERN = Pattern.compile("Publishing (build scan|build information)\\.\\.\\.");
+    private static final Pattern URL_PATTERN = Pattern.compile(".*(?:\\[INFO] )?(https?://.*/s/.*)");
+    private static final int LINES_TO_SCAN = 1000;
     private final BuildScanPublishedListener listener;
     private int linesSinceBuildScanPublishingMessage = Integer.MAX_VALUE;
-    private static final Pattern BUILD_SCAN_PATTERN = Pattern.compile("Publishing (build scan|build information)\\.\\.\\.");
-    private static final Pattern URL_PATTERN = Pattern.compile("https?://\\S*");
 
     public BuildScanLogScanner(BuildScanPublishedListener listener) {
         this.listener = listener;
     }
 
     void scanLine(String line) {
-        if (linesSinceBuildScanPublishingMessage < 10) {
+        if (linesSinceBuildScanPublishingMessage < LINES_TO_SCAN) {
             linesSinceBuildScanPublishingMessage++;
-            Matcher matcher = URL_PATTERN.matcher(line);
-            if (matcher.find()) {
+            Optional<String> maybeBuildScanUrl = doGetBuildScanId(line);
+            if (maybeBuildScanUrl.isPresent()) {
                 linesSinceBuildScanPublishingMessage = Integer.MAX_VALUE;
-                String buildScanUrl = matcher.group();
-                listener.onBuildScanPublished(buildScanUrl);
+                listener.onBuildScanPublished(maybeBuildScanUrl.get());
             }
         }
         if (BUILD_SCAN_PATTERN.matcher(line).find()) {
             linesSinceBuildScanPublishingMessage = 0;
         }
 
+    }
+
+    private static Optional<String> doGetBuildScanId(String text) {
+        Matcher matcher = URL_PATTERN.matcher(text);
+        return matcher.matches() ? Optional.of(matcher.group(1)) : Optional.empty();
     }
 }
--- a/src/main/java/hudson/plugins/gradle/BuildScanLogScanner.java
+++ b/src/main/java/hudson/plugins/gradle/BuildScanLogScanner.java
@@ -18,11 +18,10 @@ public class BuildScanLogScanner {
     void scanLine(String line) {
         if (linesSinceBuildScanPublishingMessage < LINES_TO_SCAN) {
             linesSinceBuildScanPublishingMessage++;
-            Optional<String> maybeBuildScanUrl = doGetBuildScanId(line);
-            if (maybeBuildScanUrl.isPresent()) {
+            tryFindBuildScanUrl(line).ifPresent(url -> {
                 linesSinceBuildScanPublishingMessage = Integer.MAX_VALUE;
-                listener.onBuildScanPublished(maybeBuildScanUrl.get());
-            }
+                listener.onBuildScanPublished(url);
+            });
         }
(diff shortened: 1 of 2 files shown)


src/test/groovy/hudson/plugins/gradle/BuildScanLogScannerTest.groovy
```diff
@@ -0,0 +1,53 @@
...
+
+        then:
+        listener.urls == expectedUrls
+
+        where:
+        log                                                                                                         || expectedUrls
+        logWithBuildScans(["https://scans.gradle.com/s/bzb4vn64kx3bc"])                                             || ["https://scans.gradle.com/s/bzb4vn64kx3bc"]
+        logWithBuildScans(["https://scans.gradle.com/s/bzb4vn64kx3bc", "https://scans.gradle.com/s/asc9wm73ly1do"]) || ["https://scans.gradle.com/s/bzb4vn64kx3bc", "https://scans.gradle.com/s/asc9wm73ly1do"]
+        logWithBuildScans(["https://scans.gradle.com/bzb4vn64kx3bc"])                                               || []
+        logWithBuildScans(["https://scans.gradle.com/bzb4vn64kx3bc"], 1010)                                         || []
```
facewindu, line 28: if the goal of this is to test that we don't grab URLS after LINES_TO_SCAN, then we should use a valid URL ("https://scans.gradle.com/s/bzb4vn64kx3bc") (with the /s/ part) otherwise it doesn't prove anything

facewindu (changes requested):

facewindu (approved):

src/test/groovy/hudson/plugins/gradle/BuildScanLogScannerTest.groovy
```diff
@@ -0,0 +1,53 @@
...
+        log.each { scanner.scanLine(it) }
+
+        then:
+        listener.urls == expectedUrls
+
+        where:
+        log                                                                                                         || expectedUrls
+        logWithBuildScans(["https://scans.gradle.com/s/bzb4vn64kx3bc"])                                             || ["https://scans.gradle.com/s/bzb4vn64kx3bc"]
+        logWithBuildScans(["https://scans.gradle.com/s/bzb4vn64kx3bc", "https://scans.gradle.com/s/asc9wm73ly1do"]) || ["https://scans.gradle.com/s/bzb4vn64kx3bc", "https://scans.gradle.com/s/asc9wm73ly1do"]
+        logWithBuildScans(["https://scans.gradle.com/bzb4vn64kx3bc"])                                               || []
```
guylabs, line 27: Can we add a test with 900 lines in between to have this case tested too ?

guylabs (approved):
Small test adaption. No need that I review again.

src/main/java/hudson/plugins/gradle/BuildScanLogScanner.java
```diff
@@ -1,31 +1,37 @@
 package hudson.plugins.gradle;
 
+import java.util.Optional;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 public class BuildScanLogScanner {
+    private static final Pattern BUILD_SCAN_PATTERN = Pattern.compile("Publishing (build scan|build information)\\.\\.\\.");
+    private static final Pattern URL_PATTERN = Pattern.compile(".*(?:\\[INFO] )?(https?://.*/s/.*)");
+    private static final int LINES_TO_SCAN = 1000;
```
c00ler, line 10: I think we would need to make it configurable in the future.

src/main/java/hudson/plugins/gradle/BuildScanLogScanner.java
```diff
@@ -1,31 +1,37 @@
...
     }
 
     void scanLine(String line) {
-        if (linesSinceBuildScanPublishingMessage < 10) {
+        if (linesSinceBuildScanPublishingMessage < LINES_TO_SCAN) {
             linesSinceBuildScanPublishingMessage++;
-            Matcher matcher = URL_PATTERN.matcher(line);
-            if (matcher.find()) {
+            Optional<String> maybeBuildScanUrl = doGetBuildScanId(line);
+            if (maybeBuildScanUrl.isPresent()) {
```
c00ler, line 22: Please, use `doGetBuildScanId(line).ifPresent(buildScanUrl -> ...)`

src/main/java/hudson/plugins/gradle/BuildScanLogScanner.java
```diff
@@ -1,31 +1,37 @@
...
+                listener.onBuildScanPublished(maybeBuildScanUrl.get());
             }
         }
         if (BUILD_SCAN_PATTERN.matcher(line).find()) {
             linesSinceBuildScanPublishingMessage = 0;
         }
 
     }
+
+    private static Optional<String> doGetBuildScanId(String text) {
```
c00ler, line 33: Maybe call it `tryFindBuildScanId(...)`, so that the name better reflects that the output might be `Optional`. WDYT?

src/test/groovy/hudson/plugins/gradle/BuildScanLogScannerTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+
+import spock.lang.Specification
+import spock.lang.Subject
+import spock.lang.Unroll
+
+@Unroll
+@Subject(BuildScanLogScanner)
+class BuildScanLogScannerTest extends Specification {
+
+    def 'properly captures build scan ur given #log'(List<String> log, List<String> expectedUrls) {
```
c00ler, line 12:
```suggestion
    def 'properly captures build scan url given #log'(List<String> log, List<String> expectedUrls) {
```

src/test/groovy/hudson/plugins/gradle/BuildScanLogScannerTest.groovy
```diff
@@ -0,0 +1,54 @@
...
+            log.add("Publishing build scan...")
+            (1..linesBetween).each { log.add("Some in-between log") }
+            log.add(url)
+            log.addAll(["Some log", "other log"])
+        }
+
+        log
+    }
+
+    class MockBuildScanPublishedListener implements BuildScanPublishedListener {
```
c00ler, line 46: I think it can be `private static` to emphasize that it's local to this test only

src/main/java/hudson/plugins/gradle/BuildScanLogScanner.java
sorin-florea: `tryFindBuildScanUrl` since it's the url and not the id that it extracts.

src/main/java/hudson/plugins/gradle/BuildScanLogScanner.java
sorin-florea: @guylabs WDYT, do we want to make it configurable and should we wait with this then?

guylabs: I'd keep that for now. If we face another issue in the future we can then always make it configurable. Not worth the efforts for now.

c00ler: Yep, no need to do it now. Something for the future.

c00ler (approved):

welandaz (approved):

PR #220, opened 2022-11-18
Add jenkins war and gradle plugin as inputs to the tests

c00ler:

### DIFF
--- a/acceptance-tests/build.gradle
+++ b/acceptance-tests/build.gradle
@@ -26,7 +26,7 @@ repositories {
 }
 
 configurations {
-    pluginsCollection {
+    gradlePlugin {
         canBeConsumed = false
         canBeResolved = true
     }
@@ -41,17 +41,7 @@ dependencies {
     testImplementation platform("io.netty:netty-bom:4.1.85.Final")
     testImplementation "io.ratpack:ratpack-test:1.9.0"
 
-    pluginsCollection project(path: ':', configuration: 'pluginsCollectionDist')
-}
-
-def testDependenciesDir = layout.buildDirectory.dir('test-dependencies')
-
-def copyTestDependencies = tasks.create("copyTestDependencies", Copy) {
-    from configurations.pluginsCollection
-    into testDependenciesDir
-
-    include '*.hpi'
-    include '*.jpi'
+    gradlePlugin project(path: ':', configuration: 'gradlePluginJpi')
 }
 
 def currentJava = JavaVersion.current()
@@ -68,8 +58,13 @@ def allTestTasks =
             jenkinsVersion.default || currentJava.isCompatibleWith(jenkinsVersion.requiredJavaVersion)
         }
         .collect { jenkinsVersion ->
-            def jenkinsWarLocation =
-                new File("${project.gradle.gradleUserHomeDir}/jenkins-cache/${jenkinsVersion.version}/jenkins.war")
+            def downloadJenkinsTask =
+                tasks.register("download_jenkins_${jenkinsVersion.label}", Download) {
+                    src jenkinsVersion.downloadUrl
+                    dest file("${project.gradle.gradleUserHomeDir}/jenkins-cache/${jenkinsVersion.version}/jenkins.war")
+                    onlyIfModified true
+                    tempAndMove true
+                }
 
             def testTask =
                 jenkinsVersion.default
@@ -77,30 +72,25 @@ def allTestTasks =
                     : tasks.register("test_${jenkinsVersion.label}", Test)
 
             testTask.configure {
-                dependsOn copyTestDependencies
+                def jenkinsWar = downloadJenkinsTask.map { it.outputs.files.singleFile }
+                def gradlePlugin = configurations.gradlePlugin
+
+                inputs.files(jenkinsWar)
+                inputs.files(gradlePlugin)
 
                 onlyIf {
                     // Do not run on Windows as written here: https://github.com/jenkinsci/acceptance-test-harness/blob/master/docs/EXTERNAL.md
                     !OperatingSystem.current().isWindows()
                 }
 
-                doFirst {
-                    download.run {
-                        src jenkinsVersion.downloadUrl
-                        dest jenkinsWarLocation
-                        onlyIfModified true
-                        tempAndMove true
-                    }
-                }
-
                 // If executed on TeamCity, we need to set the Chromedriver path
                 if (gradle.ciTeamCityBuild) {
-                    systemProperty 'webdriver.chrome.driver', System.getenv('HOME') + '/.gradle/webdriver/chromedriver/chromedriver'
+                    systemProperty 'webdriver.chrome.driver', "${System.getenv('HOME')}/.gradle/webdriver/chromedriver/chromedriver"
                 }
 
                 environment([
-                    JENKINS_WAR: jenkinsWarLocation,
-                    LOCAL_JARS : testDependenciesDir.get().file("gradle.hpi"),
+                    JENKINS_WAR: jenkinsWar.get(),
+                    LOCAL_JARS : gradlePlugin.singleFile,
                     BROWSER    : gradle.ciJenkinsBuild ? 'firefox-container' : 'chrome'
                 ])
 
--- a/build.gradle
+++ b/build.gradle
@@ -68,7 +68,7 @@ java {
 
 configurations {
     includedLibs
-    pluginsCollectionDist {
+    gradlePluginJpi {
         canBeConsumed = true
         canBeResolved = false
     }
@@ -254,5 +254,5 @@ jar {
 }
 
 artifacts {
-    pluginsCollectionDist(jpi)
+    gradlePluginJpi(jpi)
 }


facewindu (approved):

guylabs (approved):

settings.gradle
```diff
@@ -45,3 +45,11 @@ buildCache {
         }
     }
 }
+
+gradle.afterProject { project ->
```
sorin-florea, line 49: Why is this executed in `afterProject`, and why not put it in `build.gradle`?

sorin-florea (changes requested):

settings.gradle
c00ler: We have multiple modules that produce an archive: root project, configuration maven extension. We need to update archiving in both modules. Maybe there will be more modules in the future that also need to have this changed.

sorin-florea: Ok, but `afterEvaluate` should cause an error in Gradle 8 and it's not compatible with configuration cache ([nr. 8](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:task_configuration_avoidance_general)). 

`afterProject` suffers from the same pitfalls given that it is [doing afterEvaluate](https://github.com/gradle/gradle/blob/d33c5d822d5269e8489fcf53057cd01fa50a0a43/subprojects/core/src/main/java/org/gradle/invocation/DefaultGradle.java#L305).

sorin-florea (approved):

settings.gradle
pbielicki: I share Sorin's concerns. `build.gradle` is a "better" place and I don't see a reason why it has to be enclosed in the deprecated `afterEvaluate` block

facewindu: you should create a convention plugin, to be applied by all projects one by one, if you don't want to duplicate this logic on all projects. You'll just have to apply your convention plugin on all projects

acceptance-tests/build.gradle
```diff
@@ -68,44 +58,46 @@ def allTestTasks =
...
 
-                environment([
-                    JENKINS_WAR: jenkinsWarLocation,
-                    LOCAL_JARS : testDependenciesDir.get().file("gradle.hpi"),
-                    BROWSER    : gradle.ciJenkinsBuild ? 'firefox-container' : 'chrome'
-                ])
+                doFirst {
+                    // If executed on TeamCity, we need to set the Chromedriver path
+                    if (gradle.ciTeamCityBuild) {
+                        systemProperty 'webdriver.chrome.driver', "${System.getenv('HOME')}/.gradle/webdriver/chromedriver/chromedriver"
```
wolfs, line 93: 🤔 Why not use a command line argument provider for doing this, so it doesn't need to happen in a `doFirst`.

acceptance-tests/build.gradle
```diff
@@ -68,44 +58,46 @@ def allTestTasks =
...
                     ? tasks.named("test")
                     : tasks.register("test_${jenkinsVersion.label}", Test)
 
             testTask.configure {
-                dependsOn copyTestDependencies
+                def jenkinsWar = downloadJenkinsTask
+                def gradlePlugin = configurations.gradlePlugin
+
+                inputs.files(jenkinsWar)
+                inputs.files(gradlePlugin)
```
wolfs, line 79: ❌ Those two inputs should have some normalization (runtime classpath?) and a property name.

build-logic/src/main/groovy/buildlogic.reproducible-archives.gradle
```diff
@@ -0,0 +1,5 @@
+// See https://docs.gradle.org/current/userguide/working_with_files.html#sec:reproducible_archives
+tasks.withType(AbstractArchiveTask).configureEach {
+    preserveFileTimestamps = false
```
wolfs, line 3: 🤔 If you use normalization, I suppose you can remove the reproducible JARs here again, right?

settings.gradle
c00ler: Great idea! Added convention plugin: https://github.com/jenkinsci/gradle-plugin/pull/220/commits/96827d1be1a07930f2eb83ac3c93eef2fb859b80

build.gradle
```diff
@@ -150,80 +153,85 @@ tasks.withType(SpotBugsTask) {
...
-    project.version
-  }
+    inputs.property('pluginVersion') {
+        project.version
+    }
+}
+
+// See https://github.com/jenkinsci/gradle-jpi-plugin/blob/master/src/main/kotlin/org/jenkinsci/gradle/plugins/manifest/GenerateJenkinsManifestTask.kt#L69
+tasks.withType(GenerateJenkinsManifestTask).configureEach {
+    version = project.version.toString().replace("-SNAPSHOT", "-dev")
```
wolfs, line 190: 🤔 I am not sure you always want to do this or only when running tests. I suppose when you upload, you don't want to use a fixed `-dev` version even for snapshot builds.

settings.gradle
c00ler: Refactored to use inputs normalization: https://github.com/jenkinsci/gradle-plugin/pull/220/commits/39c08b42578feaef5c793a1055e7d288a6faa895

acceptance-tests/build.gradle
```diff
@@ -68,44 +58,46 @@ def allTestTasks =
             jenkinsVersion.default || currentJava.isCompatibleWith(jenkinsVersion.requiredJavaVersion)
         }
         .collect { jenkinsVersion ->
-            def jenkinsWarLocation =
-                new File("${project.gradle.gradleUserHomeDir}/jenkins-cache/${jenkinsVersion.version}/jenkins.war")
+            def downloadJenkinsTask =
+                tasks.register("download_jenkins_${jenkinsVersion.label}", Download) {
```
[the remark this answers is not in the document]
c00ler: Good idea! Changed to the camel case.

acceptance-tests/build.gradle
c00ler: There is no need to have it in the `doFirst` block. I can move it outside. Are there any downsides to having it in the `doFirst` block?

acceptance-tests/build.gradle
c00ler: Added inputs normalization.

build.gradle
```diff
@@ -1,59 +1,60 @@
...
 
+import java.util.zip.ZipFile
+
 plugins {
-  id 'org.gradle.test-retry' version '1.4.1'
-  id 'org.jenkins-ci.jpi' version '0.43.0'
-  id 'ru.vyarus.animalsniffer' version '1.6.0'
-  id 'com.github.spotbugs' version '5.0.13'
-  id 'codenarc'
+    id 'org.gradle.test-retry' version '1.4.1'
```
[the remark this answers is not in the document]
c00ler: We use 4 spaces in all our projects. It's easier to switch between projects when using the same pattern everywhere. In `gradle-plugin` it was mixed already, even inside one file. Changed it to the most common convention.

welandaz (approved):

build.gradle
c00ler: Yes, in general, we don't want `-dev` to leak anywhere. At the moment we don't have use cases except testing where we use `-SNAPSHOT`. It's a temporary "hack". Hopefully, my PR: https://github.com/jenkinsci/gradle-jpi-plugin/pull/210 will be accepted and we will be able to get rid of it soon.

acceptance-tests/build.gradle
wolfs: Changing inputs in a `doFirst` block is a bad practice since it tries to circumvent Gradle's input/output detection and may cause undeclared inputs. We also want to deprecate being able to modify inputs in `doFirst` at some point.

build.gradle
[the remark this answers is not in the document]
wolfs: Please move this change to a separate PR and include a `.editorconfig` with the right settings there. With the `.editorconfig`, you can even specify 2 spaces in `*.gradle` files and IDEA will respect that.

acceptance-tests/build.gradle
```diff
@@ -68,44 +60,46 @@ def allTestTasks =
...
-                    : tasks.register("test_${jenkinsVersion.label}", Test)
+                    : tasks.register("test${jenkinsVersion.label}", Test)
 
             testTask.configure {
-                dependsOn copyTestDependencies
+                def jenkinsWar = downloadJenkinsTask
+                def gradlePlugin = configurations.gradlePlugin
+
+                inputs.files(jenkinsWar).withPropertyName("jenkinsWar")
+                inputs.files(gradlePlugin).withPropertyName("gradlePlugin").withNormalizer(ClasspathNormalizer)
```
wolfs, line 81: ❌ You need should at least select a path sensitivity if you don't use `ClasspathNormalizer`.

I think that the current declaration is turned around as well: the war file could use a `ClasspathNormalizer`, since the `.war` extension is recognized by Gradle. While the normalizer doesn't have more effect than `PathSensitivity.NONE` on `gradlePlugin`, since `.hpi` is not recognized by Gradle.

I'd add `ClasspathNormalizer` for both inputs and add a comment to `gradlePlugin` that the zip normalization doesn't work there, yet.

build.gradle
wolfs: This may have an effect for manual testing when you deploy the HPI file to your locally running Jenkins, right?

build.gradle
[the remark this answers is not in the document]
c00ler: OK. Will revert for now.

acceptance-tests/build.gradle
c00ler: Thank you for the suggestion. Applied.

acceptance-tests/build.gradle
```diff
@@ -68,44 +60,51 @@ def allTestTasks =
...
-                        tempAndMove true
-                    }
+                javaLauncher = javaToolchains.launcherFor {
+                    languageVersion = jenkinsVersion.javaVersion
                 }
 
                 // If executed on TeamCity, we need to set the Chromedriver path
                 if (gradle.ciTeamCityBuild) {
                     systemProperty 'webdriver.chrome.driver', System.getenv('HOME') + '/.gradle/webdriver/chromedriver/chromedriver'
                 }
```
wolfs, line 100: ❌ My comment was to use a `CommandLineArgumentProvider` to pass in the system property to the test task. This way you can avoid that the test task has the absolute path to the Chrome driver as an input.

c00ler: Sorry, missed this one. Moved the logic to the provider: https://github.com/jenkinsci/gradle-plugin/pull/220/commits/0ebfe53f808e0d57e4d08bc87eb101fd879241de

facewindu: nit, but I would pass `gradle.ciTeamCityBuild` as a param to not hold on `Gradle` as a field in your provider

wolfs (approved):

PR #225, opened 2022-12-13
Upgrade bundled Gradle Enterprise Maven extension version.

c00ler:
Upgrade bundled Gradle Enterprise Maven extension version to `1.16`.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -225,6 +225,7 @@ The following table shows the compatibility of the plugin version with Gradle En
 
 |===
 |Jenkins Gradle Plugin version  | Injected Gradle Enterprise Maven extension version | Injected Common Custom User Data Maven extension version  | Minimum supported Gradle Enterprise version
+|2.2-SNAPSHOT                   | 1.16                                               | 1.11.1                                                    | 2022.3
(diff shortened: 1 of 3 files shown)


facewindu (commented):
1.16.1 is coming in a few days.? Shouldn't we update to it instead?

README.adoc
```diff
@@ -225,6 +225,7 @@ The following table shows the compatibility of the plugin version with Gradle En
 
 |===
 |Jenkins Gradle Plugin version  | Injected Gradle Enterprise Maven extension version | Injected Common Custom User Data Maven extension version  | Minimum supported Gradle Enterprise version
+|2.2-SNAPSHOT                   | 1.16                                               | 1.11.1                                                    | 2022.3
```
guylabs, line 228: I'd rather use something like:
```suggestion
|Next version                   | 1.16                                               | 1.11.1                                                    | 2022.3
```

guylabs (approved):
Similar to what Fracois said, let's use `1.16.1` once released and then do another release of the Jenkins plugin.

c00ler:
@facewindu @guylabs Good point about the next version. I will keep the PR open until `1.16.1` is released and then just change `1.6 -> 1.6.1`.

sorin-florea (approved):

PR #226, opened 2022-12-14
Migrate build scripts to Kotlin DSL

c00ler:
Migrate the build logic to use Kotlin DSL. To avoid problems, I tried to keep the build logic as close as possible to the Groovy variant. Later on, we might think about some refactorings.


guylabs (approved):

welandaz (approved):

sorin-florea (approved):

build.gradle.kts
```diff
@@ -0,0 +1,286 @@
...
+    from(includedLibs) {
+        into("hudson/plugins/gradle/injection")
+    }
+    from(generateVersionFile.outputs.files) {
+        into("versions")
+    }
+}
+
+artifacts {
+    add(gradlePluginJpi.name, jpi)
```
wolfs, line 285:
```suggestion
    gradlePluginJpi(jpi)
```

Does this work?

wolfs (approved):
Nice, thank you!

build.gradle.kts
c00ler: No, unfortunately, it's not possible in Kotlin DSL.

PR #232, opened 2023-01-04
Enriched build scans

jprinet:
Enrich Jenkins job summary page with the following build scan data fetched from Gradle Enterprise API:
- project name
- requested tasks
- build tool version
- build outcome
- build scan links

The feature is turned off by default and can be enabled and configured through the Jenkins system administration.

Worth noticing that the `InjectionConfig` has been moved to `GlobalConfig`


src/test/groovy/hudson/plugins/gradle/enriched/ScanDetailTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+    def 'Get scan detail with HTTP error on first request'() {
+        given:
+        def httpClientProvider = Stub(HttpClientProvider)
+        def enrichedSummaryConfig = Stub(EnrichedSummaryConfig)
+        def scanDetailService = new ScanDetailServiceDefaultImpl(Secret.fromString("{c2VjcmV0}"), new URI("https://foo.bar"))
+        scanDetailService.httpClientProvider = httpClientProvider
+        scanDetailService.enrichedSummaryConfig = enrichedSummaryConfig
+        enrichedSummaryConfig.isEnabled() >> true
+        CloseableHttpClient httpClient = Stub(CloseableHttpClient)
+        def response1 = Stub(CloseableHttpResponse)
```
c00ler, line 51:
```suggestion
        def response = Stub(CloseableHttpResponse)
```

src/test/groovy/hudson/plugins/gradle/enriched/ScanDetailTest.groovy
```diff
@@ -0,0 +1,153 @@
...
+        response1.getStatusLine() >> new BasicStatusLine(HttpVersion.HTTP_1_1, HttpStatus.SC_OK, "")
+        response1.getEntity() >> new StringEntity("{\"buildToolType\":\"" + buildToolType + "\",\"buildToolVersion\":\"7.5.1\"}")
```
c00ler, line 74: We can make the response a bit easier to read by using multiline strings:

```
new StringEntity(
    """
    {
        "buildToolType": "${buildToolType}",
        "buildToolVersion": "7.5.1"   
    }
    """.stripIndent())
```

src/main/java/hudson/plugins/gradle/config/GlobalConfig.java
```diff
@@ -20,31 +22,43 @@
...
+                    "JENKINSGRADLEPLUGIN_CCUD_PLUGIN_VERSION",
+                    "JENKINSGRADLEPLUGIN_GRADLE_PLUGIN_REPOSITORY_URL",
+                    "JENKINSGRADLEPLUGIN_GRADLE_INJECTION_ENABLED_NODES",
+                    "JENKINSGRADLEPLUGIN_GRADLE_INJECTION_DISABLED_NODES",
+                    "JENKINSGRADLEPLUGIN_GRADLE_ENTERPRISE_EXTENSION_VERSION",
+                    "JENKINSGRADLEPLUGIN_CCUD_EXTENSION_VERSION",
+                    "JENKINSGRADLEPLUGIN_MAVEN_INJECTION_ENABLED_NODES",
+                    "JENKINSGRADLEPLUGIN_MAVEN_INJECTION_DISABLED_NODES"
+            );
+    private boolean enrichedSummaryEnabled;
```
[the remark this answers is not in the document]
c00ler: This comment will probably be resolved automatically after we discuss [this](https://github.com/jenkinsci/gradle-plugin/pull/232#pullrequestreview-1251938011) 😄

README.adoc
```diff
@@ -87,6 +87,29 @@ The `withGradle` wrapper should be used instead, since it also deals well with p
 
 image::find-build-scans.png[Find build scans,{thumbnail}]
 
+== Enriched build scans
```
c00ler, line 90: ❓ Should we also mention that it will work in both cases if the user has configured GE manually or the auto-injection was used?

src/test/groovy/hudson/plugins/gradle/enriched/ScanDetailServiceDefaultImplTest.groovy
```diff
@@ -0,0 +1,212 @@
...
+        scanDetail.url == "https://foo.bar/s/scanId"
+        scanDetail.buildToolType.toString() == buildToolType.toUpperCase()
+        scanDetail.buildToolVersion == "7.5.1"
+        scanDetail.projectName == "project"
+        scanDetail.tasks == [ "clean", "build" ]
+        !scanDetail.hasFailed
+
+        where:
+        buildToolType | httpResponseBody
+        "gradle"      | "{\"foo\":\"bar\",\"rootProjectName\":\"project\",\"requestedTasks\":[\"clean\",\"build\"],\"hasFailed\":false}"
```
c00ler, line 168: It is possible to use `'` or `"""` quotes to define a string, then there is no need to escape `"`. It's a bit easier to read

```suggestion
        "gradle"      | '{"foo":"bar","rootProjectName":"project","requestedTasks":["clean","build"],"hasFailed":false}'
```

src/main/java/hudson/plugins/gradle/enriched/ScanDetailService.java
```diff
@@ -0,0 +1,139 @@
...
+                        if (httpEntityApiBuildAttributes != null) {
+                            String apiBuildAttributesResponse = EntityUtils.toString(httpEntityApiBuildAttributes);
+                            scanDetail = objectReader.readValue(apiBuildAttributesResponse);
+                            return scanDetail;
+                        }
+                    }
+                }
+            }
+        } catch (Exception e) {
+            LOGGER.info("Error fetching data [{}]", e.getMessage());
```
c00ler, line 106: It's better to log with the stack trace, so we can have more info in case we need to investigate it.

```suggestion
            LOGGER.warn("Error fetching build scan details", e);
```

src/main/java/hudson/plugins/gradle/enriched/ScanDetailService.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private ScanDetail doGetScanDetail(String buildScanUrl) {
+        if (buildScanUrl.startsWith(GRADLE_ENTERPRISE_PUBLIC_SERVER)) {
+            // API is not accessible on public server
+            return null;
+        }
+
+        String baseApiUri = getBaseApiUri(buildScanUrl);
+        if (null == baseApiUri || baseApiUri.isEmpty()) {
+            LOGGER.error("Gradle Enterprise API URL can't be resolved");
```
c00ler, line 68: I think we don't need logging here. If URL parsing fails in `getBaseApiUri`, then we will already log a warning there. There is no need to call twice. WDYT?

src/main/java/hudson/plugins/gradle/enriched/ScanDetailService.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+        String scanId = buildScanUrl.substring(scanIdStartIndex + URL_CONTEXT_PATH_SCAN_ID.length());
+
+        try {
+            URI baseApiUri = buildScanServer != null ?
+                    URI.create(buildScanServer)
+                    : URI.create(buildScanUrl).resolve("/");
+
+            return baseApiUri.resolve(URL_CONTEXT_PATH_API_BUILDS).resolve(scanId).toASCIIString();
+        } catch(IllegalArgumentException e) {
```
c00ler, line 125:
```suggestion
        } catch (IllegalArgumentException e) {
```

src/main/java/hudson/plugins/gradle/AbstractBuildScanAction.java
```diff
@@ -40,11 +43,22 @@ public void addScanUrl(String scanUrl) {
         }
     }
 
+    public void addScanDetail(ScanDetail scanDetail) {
+        if (!scanDetails.contains(scanDetail)) {
```
c00ler, line 47: In order for this check to work we would need to override `equals` method of the `ScanDetail` class. Alternatively, I think we can use a `Map<String, ScanDetail>`, where the key is a scan URL. Then we can easily deduplicate by the URL.

c00ler: Could you also please add a simple unit test to ensure that we can't add duplicates?

src/main/java/hudson/plugins/gradle/enriched/ScanDetailService.java
jprinet: Agreed, I added a log message in `getBaseUri` when buildScanId can't be parsed to get fully covered

src/main/java/hudson/plugins/gradle/AbstractBuildScanAction.java
jprinet: Well spotted 👍
I added a unit test.

c00ler (approved):

PR #243, opened 2023-02-10
Update gradle to version 8.0.1

c00ler:


guylabs (approved):

settings.gradle.kts
```diff
@@ -1,6 +1,7 @@
 plugins {
     id("com.gradle.enterprise") version "3.12.3"
     id("com.gradle.common-custom-user-data-gradle-plugin") version "1.8.2"
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.4.0"
```
guylabs, line 4: Why is that required for the Gradle upgrade?

gradle.properties
```diff
@@ -1,2 +1,3 @@
 version = 2.3-SNAPSHOT
 githubUser = jenkinsci
+org.gradle.jvmargs=-Xmx1g -XX:MaxMetaspaceSize=512m
```
guylabs, line 3: Keep in mind that this is also applied to any CI agents where this project is running on. Does it mean that with Gradle 8 you get any kind of OOM errors?

welandaz (approved):

sorin-florea (approved):

settings.gradle.kts
c00ler: I was also surprised, but auto-discovery was not working properly on jenkins: https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fgradle-plugin/detail/PR-243/6/pipeline/21

gradle.properties
c00ler: Yes, exactly, with the standard memory limits it was not working on jenkins.

PR #258, opened 2023-03-12
Do not set empty MAVEN_OPTS environment variable

c00ler:
This pull request introduces the following changes:
* When maven auto-injection is disabled, update `MAVEN_OPTS` only if it has the changes added by auto-injection
* When maven auto-injection is disabled, remove `MAVEN_OPTS` instead of setting an empty value
* When maven auto-injection is enabled, merge changes with existing `MAVEN_OPTS` defined in the node properties. 

Fixes https://issues.jenkins.io/browse/JENKINS-70663 and https://issues.jenkins.io/browse/JENKINS-70692

### DIFF
--- a/src/main/java/hudson/plugins/gradle/injection/MavenOptsSetter.java
+++ b/src/main/java/hudson/plugins/gradle/injection/MavenOptsSetter.java
@@ -1,7 +1,6 @@
 package hudson.plugins.gradle.injection;
 
 import hudson.model.Node;
-import org.apache.commons.lang3.StringUtils;
 
 import javax.annotation.Nullable;
 import java.io.IOException;
@@ -51,7 +50,7 @@ class MavenOptsSetter {
 
     void removeIfNeeded(Node node) throws IOException, InterruptedException {
         String currentMavenOpts = getCurrentMavenOpts(node);
-        if (StringUtils.isBlank(currentMavenOpts)) {
+        if (currentMavenOpts == null || currentMavenOpts.isEmpty()) {
             return;
         }
 
--- a/src/main/java/hudson/plugins/gradle/injection/MavenOptsSetter.java
+++ b/src/main/java/hudson/plugins/gradle/injection/MavenOptsSetter.java
@@ -1,11 +1,12 @@
 package hudson.plugins.gradle.injection;
 
-import com.google.common.collect.ImmutableSet;
-import hudson.EnvVars;
 import hudson.model.Node;
+import org.apache.commons.lang3.StringUtils;
 
+import javax.annotation.Nullable;
 import java.io.IOException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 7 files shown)


src/main/java/hudson/plugins/gradle/injection/MavenOptsSetter.java
```diff
@@ -17,54 +17,81 @@ class MavenOptsSetter {
...
+        String mavenOpts = filtered(currentMavenOpts).orElse(null);
         EnvUtil.setEnvVar(node, MAVEN_OPTS_VAR, mavenOpts);
     }
 
-    private String getMavenOpts(Node node) throws IOException, InterruptedException {
-        EnvVars nodeEnvVars = EnvVars.getRemote(node.getChannel());
-        return nodeEnvVars.get(MAVEN_OPTS_VAR);
+    @Nullable
+    private static String getCurrentMavenOpts(Node node) {
+        return EnvUtil.getEnv(node, MAVEN_OPTS_VAR);
```
alextu, line 68: What about changing `EnvUtil.getEnv` to return an `Optional` directly ?

src/main/java/hudson/plugins/gradle/injection/MavenOptsSetter.java
```diff
@@ -17,54 +17,81 @@ class MavenOptsSetter {
...
     }
 
-    private String removeSystemProperties(String mavenOpts) throws RuntimeException {
+    private Optional<String> filtered(@Nullable String mavenOpts) throws RuntimeException {
         return Optional.ofNullable(mavenOpts)
-            .map(this::filterMavenOpts)
-            .orElse("");
+            .map(this::filterMavenOpts);
```
alextu, line 73: Same, how about changing the return result of `filterMavenOpts` by `Optional<String>` and `flatMap` here ?

alextu (approved):

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
```diff
@@ -109,6 +110,70 @@ class BuildScanInjectionMavenIntegrationTest extends BaseJenkinsIntegrationTest
...
+
+        withNodeEnvVars(agent) {
+            put('MAVEN_OPTS', mavenOpts)
+        }
+
+        when:
+        turnOnBuildInjectionAndRestart(agent)
+
+        then:
+        with(getMavenOptsFromNodeProperties(agent).split(" ").iterator()) {
```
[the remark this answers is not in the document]
c00ler: Yes, both approaches are applicable. I find the iterator approach more expressive and more flexible. It enforces the order, without being dependent on a particular collection type.

src/main/java/hudson/plugins/gradle/injection/MavenOptsSetter.java
c00ler: It's possible. But I'm not sure about the value though. This is a private method with a very limited scope. It's quite easy to ensure that `null` value is properly handled. Not all usages will really benefit from an Optional here. See `removeIfNeeded` method for example, where we would need to resolve optional straightaway to be able to perform a check.

welandaz (approved):

PR #269, opened 2023-03-27
Move injection logic to EnvironmentContributor

welandaz:
This work moves part of the injection logic from current `ComputerListener` call to specific `EnvironmentContributor` classes and, hence, making it a 2 step process:
1) `GradleBuildScanInjection` and `MavenBuildScanInjection` called from `ComputerListener` deal with node based injection and only copy the necessary files (init script for Gradle and extension JARs for Maven). `MavenBuildScanInjection` set's intermediate environment variables - JENKINSGRADLEPLUGIN_MAVEN_OPTS_PREPARED and JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_EXT_CLASSPATH_PREPARED. These will be used later to make a decision about enabling/disabling injection for Maven.
2) 3 new `RunListener` implementation are introduced for Maven builds. These will add `Action` to the `Run` instance with contents of those previously set variables to be used in the next step by `EnvironmentContributor` (this is needed as `EnvironmentContributor` doesn't see environment variables set in (1) for Workflow builds, only FreeStyle ones)
2) 2 new `EnvironmentContributor` implementations are introduced - `GradleInjectionEnvironmentContributor` and `MavenInjectionEnvironmentContributor`. These will provision necessary environment variables for the duration of the job run such that injection is enabled (e.g. MAVEN_OPTs/Maven plugin specific ones for Maven builds and new `GRADLE_ENTERPRISE_GRADLE_INJECTION_ENABLED` flag that enables init script). 

As a result, it would be possible to add job based injection logic in future, where we can observe job details provided by an `EnvironmentContributor` and decide whether we want to perform the injection or not. The node based injection still takes place in step (1), since we don't want to copy unnecessary files to the nodes we know should be ignored right away.

We still need to keep the cleanup code in place to remove the previously set environment variables by the old plugin versions. Those will be removed when injection will be toggled on/off and not set on the machine anymore.


src/main/java/hudson/plugins/gradle/injection/GradleBuildScanInjection.java
```diff
@@ -173,6 +107,9 @@ private boolean initScriptChanged(FilePath gradleInitScriptFile) throws IOExcept
     private void cleanup(Node node, String initScriptDirectory) {
         try {
             removeInitScript(node.getChannel(), initScriptDirectory);
+
+            // we still need to clean up all environment variables set in older
+            // versions even though we now set them in EnvironmentContributor
```
c00ler, line 112: Maybe just add that this needs to be removed in the future. This is a temporary behavior.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvironmentContributor.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static hudson.plugins.gradle.injection.MavenOptsHandler.MAVEN_OPTS;
+
+@Extension
+public class MavenInjectionEnvironmentContributor extends EnvironmentContributor implements MavenInjectionAware {
+
+    private static final MavenOptsHandler MAVEN_OPTS_HANDLER = new MavenOptsHandler(
```
c00ler, line 24: I think it can be moved to the `MavenInjectionAware` and shared as well.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvironmentContributor.java
```diff
@@ -0,0 +1,105 @@
...
+
+        // Configuration needed to support https://plugins.jenkins.io/maven-plugin/
+        extensions.add(extensionsHandler.getExtensionLocation(MavenExtensionsHandler.MavenExtension.CONFIGURATION, nodeRootPath));
+
+        envs.put(JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_EXT_CLASSPATH, constructExtClasspath(extensions, isUnix));
+        envs.put(JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_SERVER_URL, server);
+        if (config.isAllowUntrusted()) {
+            envs.put(JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_ALLOW_UNTRUSTED_SERVER, "true");
+        } else {
+            envs.remove(JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_ALLOW_UNTRUSTED_SERVER);
```
c00ler, line 84: Is it for backward compatibility? Let's add a comment.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvironmentContributor.java
```diff
@@ -0,0 +1,105 @@
...
+        List<SystemProperty> systemProperties = new ArrayList<>();
+        systemProperties.add(new SystemProperty(MAVEN_EXT_CLASS_PATH_PROPERTY_KEY, constructExtClasspath(extensions, isUnix)));
+        systemProperties.add(new SystemProperty(BUILD_SCAN_UPLOAD_IN_BACKGROUND_PROPERTY_KEY, "false"));
+
+        systemProperties.add(new SystemProperty(GRADLE_ENTERPRISE_URL_PROPERTY_KEY, server));
+        if (config.isAllowUntrusted()) {
+            systemProperties.add(new SystemProperty(GRADLE_ENTERPRISE_ALLOW_UNTRUSTED_SERVER_PROPERTY_KEY, "true"));
+        }
+
+        envs.put(MAVEN_OPTS, MAVEN_OPTS_HANDLER.create(envs, systemProperties));
```
c00ler, line 74: Maybe call the method `merge`?

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
```diff
@@ -137,109 +132,49 @@ class BuildScanInjectionMavenIntegrationTest extends BaseJenkinsIntegrationTest
...
+        extensionDirectoryEmpty.list().find { it.name == CONFIGURATION_EXTENSION_JAR } == null
+        extensionDirectoryEmpty.list().find { it.name == CCUD_EXTENSION_JAR } == null
+
         noMavenOpts(slave)
     }
 
     def 'build scan is published without GE plugin with simple pipeline'() {
         given:
-        createSlaveAndTurnOnInjection()
+        def slave= createSlaveAndTurnOnInjection()
```
c00ler, line 177:
```suggestion
        def slave = createSlaveAndTurnOnInjection()
```

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
c00ler, line 1: Is there a test for the case when previously set MAVEN_OPTS is actually deleted when we disable injection? The backward compatibility use-case.

src/main/java/hudson/plugins/gradle/injection/MavenBuildScanInjection.java
```diff
@@ -113,41 +65,18 @@ private void inject(Node node, FilePath nodeRootPath) {
             EnvUtil.setEnvVar(node, JENKINSGRADLEPLUGIN_MAVEN_AUTO_INJECTION, "true");
```
alextu, line 65: I don't see this env var being read elsewhere, what is it for ?

src/main/resources/hudson/plugins/gradle/injection/init-script.gradle
```diff
@@ -14,6 +14,11 @@ initscript {
         return System.getProperty(name) ?: System.getenv(envVarName)
     }
 
+    // finish early if injection is disabled
+    if (getInputParam("GRADLE_ENTERPRISE_GRADLE_INJECTION_ENABLED") == null) {
```
alextu, line 18: Env var name is not consistent with others as well as the call to the `getInputParam`. Maybe we should just change the initscript to only env var names are read ? It's easier to follow when looking for env var usage.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvironmentContributor.java
```diff
@@ -0,0 +1,105 @@
...
+
+        if (!isInjectionEnabled(node)) {
+            return;
+        }
+
+        InjectionConfig config = InjectionConfig.get();
+        String server = config.getServer();
+
+        List<FilePath> extensions = new LinkedList<>();
+        extensions.add(extensionsHandler.getExtensionLocation(MavenExtensionsHandler.MavenExtension.GRADLE_ENTERPRISE, nodeRootPath));
```
alextu, line 58: I'm a bit worried by the cost of calling `getExtensionLocation` since it's doing IO (maybe remote IO ?) and this will be called many times. Maybe let's be on the safe side and follow the Jenkins javadoc:

> This method can be called repeatedly for the same Run, thus the computation of this method needs to be efficient. If you have a time-consuming computation, one strategy is to take the hit once and then add the result as InvisibleAction to Run

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvironmentContributor.java
welandaz: Yup, overlooked this one; moved

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvironmentContributor.java
```diff
@@ -0,0 +1,105 @@
...
+
+        Node node = executor.getOwner().getNode();
+        if (node == null) {
+            return;
+        }
+
+        FilePath nodeRootPath = node.getRootPath();
+        if (nodeRootPath == null) {
+            return;
+        }
```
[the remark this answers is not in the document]
welandaz: Yeah, seems like `node.getRootPath();` isn't that cheap, so for cases where injection is simply off it's faster to check that first.

Done.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvironmentContributor.java
welandaz: This is actually not needed, as if the `config.isAllowUntrusted` is true, we will set it here as well (which won't clash with the env var on the node). And if it is disabled, that means that we don't need to do anything here, since it was already like that or it was turned off in which case our cleanup part in `MavenBuildScanInjection` deletes it on configuration save.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvironmentContributor.java
welandaz: Agree, sounds better.

Done.

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
welandaz: Added a space.

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
welandaz: There wasn't exactly that one, so I added `delete MAVEN_OPTS and maven-plugin variables set by older versions if auto-injection is disabled`

src/main/resources/hudson/plugins/gradle/injection/init-script.gradle
welandaz: I initially wanted it to be it's own separate thing and not bind it to Jenkins config as its not something user can influence. But I agree that it looks a bit odd in the init script (and in the code too for that matter) and we already had a similar example of `JENKINSGRADLEPLUGIN_GRADLE_AUTO_INJECTION`, so I renamed it to make the name consistent.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvironmentContributor.java
welandaz: Are you talking about the `FilePath.child(...)` call underneath?

c00ler (commented):
@welandaz Nice work! We might try to improve `MavenInjectionEnvironmentContributor`, by computing all the needed changes once and then saving them into the `InvisibleAction` (only if the injection is enabled). On the next iteration, instead of computing everything again, we can just get the precomputed values from the action and set them as env vars. @alextu WDYT?

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
```diff
@@ -0,0 +1,96 @@
...
+            }
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Action that holds Maven properties to be set in EnvironmentContributor
+     */
+    class PreparedMavenProperties extends InvisibleAction {
```
alextu, line 85: I miss something in the name that indicates it's specific to a Node (the only reason we had to setup all this machinery). So maybe something like `NodePreparedMavenInfos` or even `NodePreparedMavenEnvsForwarder` ?

welandaz: Agree, I renamed it to `NodePreparedMavenEnvsAction` (most of other actions have this convention in naming).

welandaz:
> > @welandaz Nice work! We might try to improve `MavenInjectionEnvironmentContributor`, by computing all the needed changes once and then saving them into the `InvisibleAction` (only if the injection is enabled). On the next iteration, instead of computing everything again, we can just get the precomputed values from the action and set them as env vars. @alextu WDYT?
> 
> Yes but perhaps we should use another action, or as I suggested in a comment to clearly indicate the responsibility of this action, which is a `Communicator` between the EnvContributor and node specific run. Then implement a `isConsumed()` or something to indicate the infos have been read.

Do you think we still need to do it though? Our current environment contributors don't do any heavy computation, they simply put values into EnvVars map.

alextu (approved):

src/main/java/hudson/plugins/gradle/injection/MavenAbstractBuildRunListener.java
```diff
@@ -0,0 +1,18 @@
...
+import hudson.maven.AbstractMavenBuild;
+import hudson.model.TaskListener;
+import hudson.model.listeners.RunListener;
+
+@Extension
+public class MavenAbstractBuildRunListener extends RunListener<AbstractMavenBuild<?, ?>> implements MavenInjectionAware {
+
+    @Override
+    public void onStarted(AbstractMavenBuild abstractMavenBuild, TaskListener listener) {
+        super.onInitialize(abstractMavenBuild);
```
c00ler, line 13: This is not needed, as we are not doing any decoration of the existing runner. All runners are invoked sequentially by Jenkins. `RunListener` is an abstract class with an empty implementation.

src/main/java/hudson/plugins/gradle/injection/MavenBuildScanInjection.java
```diff
@@ -117,7 +67,7 @@ private void inject(Node node, FilePath nodeRootPath) {
 
             LOGGER.info("Injecting Maven extensions " + nodeRootPath);
 
-            List<FilePath> extensions = new LinkedList<>();
+            List<FilePath> extensions = new ArrayList<>();
```
c00ler, line 70: 👍

src/main/java/hudson/plugins/gradle/injection/MavenBuildScanInjection.java
```diff
@@ -136,18 +86,12 @@ private void inject(Node node, FilePath nodeRootPath) {
                 systemProperties.add(new SystemProperty(GRADLE_ENTERPRISE_ALLOW_UNTRUSTED_SERVER_PROPERTY_KEY, "true"));
             }
 
-            MAVEN_OPTS_SETTER.append(node, systemProperties);
+            EnvUtil.setEnvVar(node, JENKINSGRADLEPLUGIN_MAVEN_OPTS_PREPARED, MAVEN_OPTS_HANDLER.merge(node, systemProperties));
```
c00ler, line 89: I don't think that `PREPARED` suffix brings any value here since we already have our prefix, which is enough for Maven to ignore it. Maybe just `JENKINSGRADLEPLUGIN_MAVEN_OPTS`?

src/main/java/hudson/plugins/gradle/injection/MavenFreeStyleRunListener.java
```diff
@@ -0,0 +1,18 @@
...
+import hudson.model.FreeStyleBuild;
+import hudson.model.TaskListener;
+import hudson.model.listeners.RunListener;
+
+@Extension
+public class MavenFreeStyleRunListener extends RunListener<FreeStyleBuild> implements MavenInjectionAware {
+
+    @Override
+    public void onStarted(FreeStyleBuild freeStyleBuild, TaskListener listener) {
+        super.onStarted(freeStyleBuild, listener);
```
c00ler, line 13: Same as in the other listener.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
```diff
@@ -0,0 +1,96 @@
...
+
+import java.io.IOException;
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public interface MavenInjectionAware {
+
+    String JENKINSGRADLEPLUGIN_MAVEN_OPTS_PREPARED = "JENKINSGRADLEPLUGIN_MAVEN_OPTS_PREPARED";
+    String JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_EXT_CLASSPATH_PREPARED = "JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_EXT_CLASSPATH_PREPARED";
```
c00ler, line 18: Same as with `JENKINSGRADLEPLUGIN_MAVEN_OPTS`. I don't think that `_PREPARED` does bring any value here.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
```diff
@@ -0,0 +1,96 @@
...
+        try {
+            EnvVars environment = run.getEnvironment(listener);
+
+            String preparedMavenOpts = environment.get(JENKINSGRADLEPLUGIN_MAVEN_OPTS_PREPARED);
+            String preparedMavenPluginConfigExtClasspath = environment.get(JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_EXT_CLASSPATH_PREPARED);
+            if (preparedMavenOpts != null && preparedMavenPluginConfigExtClasspath != null) {
+                run.addAction(new NodePreparedMavenEnvsAction(preparedMavenOpts, preparedMavenPluginConfigExtClasspath));
+            }
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException(e);
```
c00ler, line 78: Are we going to break the build if it happens? I think we should avoid doing so. Skipping the injection and just logging it, should be enough.

src/main/java/hudson/plugins/gradle/injection/MavenFreeStyleRunListener.java
```diff
@@ -0,0 +1,18 @@
...
+import hudson.Extension;
+import hudson.model.FreeStyleBuild;
+import hudson.model.TaskListener;
+import hudson.model.listeners.RunListener;
+
+@Extension
+public class MavenFreeStyleRunListener extends RunListener<FreeStyleBuild> implements MavenInjectionAware {
+
+    @Override
+    public void onStarted(FreeStyleBuild freeStyleBuild, TaskListener listener) {
```
c00ler, line 12: Is it guaranteed that it's called only once?

src/main/java/hudson/plugins/gradle/injection/MavenAbstractBuildRunListener.java
```diff
@@ -0,0 +1,18 @@
...
+
+        setPreparedMavenProperties(abstractMavenBuild, listener);
```
c00ler, line 15: Why this method is located in `MavenInjectionAware`? This logic is very specific to the `RunListener`. I don't think that we want it to leak into the shared code. If you want to share it between different listeners, then please create a custom abstract listener that can be used to share that code.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+
+    /**
+     * Action that holds Maven environment variables to be set in EnvironmentContributor
+     */
+    class NodePreparedMavenEnvsAction extends InvisibleAction {
```
c00ler, line 85: Maybe call it `MavenInjectionEnvVarsAction` to make the name more aligned with the env contributor?

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
```diff
@@ -0,0 +1,96 @@
...
+            if (preparedMavenOpts != null && preparedMavenPluginConfigExtClasspath != null) {
+                run.addAction(new NodePreparedMavenEnvsAction(preparedMavenOpts, preparedMavenPluginConfigExtClasspath));
+            }
+        } catch (IOException | InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Action that holds Maven environment variables to be set in EnvironmentContributor
```
c00ler, line 83: I think we need to reference a concrete contributor in the javadoc, not the abstract contributor, as this is a very specific action:

```suggestion
     * Action that holds Maven environment variables to be set in {@link MavenInjectionEnvironmentContributor}.
```

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
```diff
@@ -0,0 +1,96 @@
...
+     */
+    class NodePreparedMavenEnvsAction extends InvisibleAction {
+
+        public final String preparedMavenOpts;
```
c00ler, line 87: These are the real maven opts that will be set to the environment right? Why not call it `mavenOpts`?

c00ler (changes requested):

c00ler:
> > > @welandaz Nice work! We might try to improve `MavenInjectionEnvironmentContributor`, by computing all the needed changes once and then saving them into the `InvisibleAction` (only if the injection is enabled). On the next iteration, instead of computing everything again, we can just get the precomputed values from the action and set them as env vars. @alextu WDYT?
> > 
> > 
> > Yes but perhaps we should use another action, or as I suggested in a comment to clearly indicate the responsibility of this action, which is a `Communicator` between the EnvContributor and node specific run. Then implement a `isConsumed()` or something to indicate the infos have been read.
> 
> Do you think we still need to do it though? Our current environment contributors don't do any heavy computation, they simply put values into EnvVars map.

No, it is not needed anymore, since the approach has changed now.

src/main/java/hudson/plugins/gradle/injection/MavenAbstractBuildRunListener.java
welandaz: I agree, that it's not needed, but I've found one or two implementations that do have it (and don't modify anything as well) and I thought keeping if it does something in future. 

Do you feel strongly about not having it here?

P.S. The invoked one should actually be `onStarted`, forgot to change it, but this is irrelevant if we remove it.

src/main/java/hudson/plugins/gradle/injection/MavenFreeStyleRunListener.java
welandaz: The documentation doesn't mentioned it, but by looking at the code and logs of job execution, I didn't find instances of it being called more than once. 
We can add a check for an existing `NodePreparedMavenEnvsAction` object in `Run` instance and not add it if exists.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
welandaz: How about `MavenInjectionNodeEnvVarsAction`? Or do you think "Node" is redundant then?

src/main/java/hudson/plugins/gradle/injection/MavenAbstractBuildRunListener.java
welandaz: I also thought about creating a separate abstract class for this, but then thought it might be too much and since we already have a Maven injection specific interface, it made sense to keep it there (it also uses variables from it, so everything maven injection related is in one place).

But perhaps you're right and extracting it would be better. I'll update this.

src/main/java/hudson/plugins/gradle/injection/MavenBuildScanInjection.java
welandaz: For MAVEN_OPTS yes, this suffix is redundant, but we already have JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_EXT_CLASSPATH variable and need another one to distinguish between them. 
So I just kept the naming of these aligned.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
welandaz: See my comment above

src/main/java/hudson/plugins/gradle/injection/MavenAbstractBuildRunListener.java
c00ler: Since this call doesn't do anything, I think it's a noise and we need to remove it.

src/main/java/hudson/plugins/gradle/injection/MavenBuildScanInjection.java
c00ler: Why you can't reuse the same variable name? We do copying exactly for that reason, that the initially set variable is not available in the context where we need it, otherwise, nothing needs to be done.

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
c00ler: I think it's redundant. We are just transferring environment variables needed for the Maven auto-injection. It makes this data to be quite generic and not bound to any node.

src/main/java/hudson/plugins/gradle/injection/MavenAbstractBuildRunListener.java
welandaz: Hah, we already extend EnvironmentContributor, so cannot extend another one here :)

So either that interface or a util class

src/main/java/hudson/plugins/gradle/injection/MavenAbstractBuildRunListener.java
welandaz: Refactored to one generic listener without this cal

src/main/java/hudson/plugins/gradle/injection/MavenBuildScanInjection.java
welandaz: As discussed, we need both of these and I think it's clearer to just set 2 separate variables instead of one and do splitting afterwards

src/main/java/hudson/plugins/gradle/injection/MavenFreeStyleRunListener.java
welandaz: Done

src/main/java/hudson/plugins/gradle/injection/MavenFreeStyleRunListener.java
welandaz: I added a check to verify Action object is not in the run execution

src/main/java/hudson/plugins/gradle/injection/MavenAbstractBuildRunListener.java
welandaz: This is now extracted into one listener implementation

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
welandaz: Done

src/main/java/hudson/plugins/gradle/injection/MavenInjectionAware.java
c00ler: Yes, it's correct. It contains node-specific data, but the contributor (the receiver of the data) doesn't have any node-specific logic. It just gets the data (whatever this data is) and sets the required variables. I see this action as a generic envelope to just transfer data from the producer to the consumer. It's essentially the same as `EnvironmentContributingAction`, just tailored to transfer only specific variables (in form of fields).

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvVarsRunListener.java
```diff
@@ -0,0 +1,64 @@
...
+import hudson.model.listeners.RunListener;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+
+import java.io.IOException;
+
+import static hudson.plugins.gradle.injection.MavenInjectionAware.JENKINSGRADLEPLUGIN_MAVEN_OPTS_PREPARED;
+import static hudson.plugins.gradle.injection.MavenInjectionAware.JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_EXT_CLASSPATH_PREPARED;
+
+@Extension
+public class MavenInjectionEnvVarsRunListener extends RunListener<Run<?, ?>> {
```
c00ler, line 19: Maybe rename it to `MavenInjectionRunListener`? To expose fewer implementation details and to be more aligned with the `MavenInjectionEnvironmentContributor` in terms of naming.

c00ler: The purpose can be described in the javadoc.

c00ler (approved):

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvVarsRunListener.java
welandaz: Done

PR #275, opened 2023-04-03
Add Git VCS repository filtering for auto-injection

welandaz:
This PR adds an ability to perform injection based on the VCS pattern. If specified, the auto-injection will apply to the VCS repositories that match the specified pattern. The pattern itself is a comma-separated list of keywords and the match is done via contains operation: if a repository URL contains word(s) specified in a filter - the auto-injection will apply. 
The feature supports both FreeStyle and Pipeline builds and in case there are multiple repositories being checked out during the job, will enable injection if at least one of them matches the filter.

Only Git VCS is currently supported and therefore, feature requires Jenkins Git Plugin to be installed (if not installed, users will see a warning message in the configuration)
Screenshot 2023-04-03 at 16 59 00


src/main/java/hudson/plugins/gradle/injection/InjectionConfig.java
```diff
@@ -223,6 +240,24 @@ public void setMavenInjectionDisabledNodes(List<NodeLabelItem> mavenInjectionDis
             mavenInjectionDisabledNodes == null ? null : ImmutableList.copyOf(mavenInjectionDisabledNodes);
     }
 
+    @DataBoundSetter
+    public void setInjectionVcsRepositoryPatterns(String injectionVcsRepositoryPatterns) {
+        this.injectionVcsRepositoryPatterns = Util.fixEmptyAndTrim(injectionVcsRepositoryPatterns);
+
+        this.parsedInjectionVcsRepositoryPatterns = this.injectionVcsRepositoryPatterns == null
+                ? ImmutableList.of()
+                : ImmutableList.copyOf(Arrays.stream(this.injectionVcsRepositoryPatterns.split(",")).map(String::trim).collect(Collectors.toList()));
```
welandaz, line 249: We can also do this in a `getParsedInjectionVcsRepositoryPatterns() ` getter every time it's invoked, but since it's changes only via configuration change in UI, I think it's ok to have it precomputed

src/main/java/hudson/plugins/gradle/injection/InjectionUtil.java
```diff
@@ -78,4 +87,49 @@ private static boolean isEnabled(Set<String> labels, Set<String> enabledNodes) {
...
+            SCM scm = project.getScm();
+
+            if (scm == null) {
+                return true;
+            }
+
+            return vcsRepositoryUrlMatches(config, scm);
+        }
+
+        return false;
```
welandaz, line 115: Is it worth covering this with an additional `try { ... } catch {...}` clause to avoid any potential errors? Or perhaps do this in our EnvironmentContributors (for Gradle and Maven)?

The build will fail if an error happens here for whatever reason.

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionGradleIntegrationTest.groovy
```diff
@@ -443,6 +444,96 @@ class BuildScanInjectionGradleIntegrationTest extends BaseGradleIntegrationTest
...
+
+        then:
+        if (pattern == "simple-") {
+            j.assertLogContains(MSG_INIT_SCRIPT_APPLIED, build2)
+        } else {
+            j.assertLogNotContains(MSG_INIT_SCRIPT_APPLIED, build2)
+        }
+
+        where:
+        pattern << ["simple-", "this-one-does-not-match"]
```
alextu, line 483: :thought_balloon: could be replaced by a closure param
```
        then:
        pattern == result(j, build2)

        where:
        pattern                   | result
        "simple-"                 | { j, b -> j.assertLogContains(MSG_INIT_SCRIPT_APPLIED, b) }
        "this-one-does-not-match" | { j, b -> j.assertLogNotContains(MSG_INIT_SCRIPT_APPLIED, b) }
```

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
```diff
@@ -503,6 +507,84 @@ node {
...
+""", false))
+
+        when:
+        // In this case, the job had to be executed twice for SCM to be populated (we do the same for Gradle, though
+        // for a different reason). Manual test didn't exhibit this problem, so most probably something caused by a test setup
+        j.buildAndAssertSuccess(pipelineJob)
+        def build = j.buildAndAssertSuccess(pipelineJob)
+
+        then:
+        if (pattern.contains("simple-")) {
```
alextu, line 546: 💭 could be replaced by a closure param (see other comment)

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
```diff
@@ -503,6 +507,84 @@ node {
...
+        } else {
+            j.assertLogNotContains("[INFO] The Gradle Terms of Service have not been agreed to.", build)
+        }
+
+        where:
+        pattern                                     |  mavenSetup
+        "not-found-pattern, simple-"                | "withEnv([\"PATH+MAVEN=\${tool 'mavenInstallationName'}/bin\"]) {"
+        "not-found-pattern, simple-"                | "withMaven(maven: 'mavenInstallationName') {"
+        "this-one-does-not-match, this-one-too"     | "withEnv([\"PATH+MAVEN=\${tool 'mavenInstallationName'}/bin\"]) {"
+        "this-one-does-not-match, this-one-too"     | "withMaven(maven: 'mavenInstallationName') {"
```
alextu, line 557: 💭 I guess it does not hurt to test VCS pattern + various Maven install, but wouldn't be enough to have separate coverage (one that test injection + various maven setup, one that test injection + various VCS pattern) ?

src/main/java/hudson/plugins/gradle/injection/InjectionUtil.java
```diff
@@ -78,4 +87,49 @@ private static boolean isEnabled(Set<String> labels, Set<String> enabledNodes) {
...
+    public static boolean isVcsRepositoryAllowed(Run<?, ?> run, InjectionConfig config) {
+        if (config.getParsedInjectionVcsRepositoryPatterns() == null || config.getParsedInjectionVcsRepositoryPatterns().isEmpty()) {
+            return true;
+        }
+
+        if (run instanceof WorkflowRun) {
+            Collection<? extends SCM> scms = ((WorkflowRun) run).getParent().getSCMs();
+
+            for (SCM scm : scms) {
+                if (vcsRepositoryUrlMatches(config, scm)) {
```
alextu, line 100: 💭 I guess here we're in a unpredictable case, at least without first experimenting, I would be curious to have some tests, for example a pipeline like:
- checkout not matching
- build maven
- checkout matching
- build maven

alextu (changes requested):
Great work !

README.adoc
```diff
@@ -134,6 +134,15 @@ Next set the URL of the Gradle Enterprise instance to which the build scans shou
 
 image::enable-auto-injection.png[Enable auto-injection,{thumbnail}]
 
+It is also possible to enable injection for specific repositories by specifying VCS repository patterns. These are comma-separated values, which will be used in a contains check against the repository URL.
+
+image::general-git-vcs-repository-filtering.png[Git VCS repository filtering, {thumbnail}]
+
```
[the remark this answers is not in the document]
welandaz: Hm...interesting thought...The original idea was to have it comma-separated in case lots of values are specified + it makes parsing predictable. 

@guylabs Any thoughts on this from your side? Since it's in beta anyway, we can also reevaluate and change it later I guess, right?

src/main/java/hudson/plugins/gradle/injection/InjectionUtil.java
welandaz: I did this test manually and injection was applied, but I can also document it as an automated test.

README.adoc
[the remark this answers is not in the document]
guylabs: Yes I'd keep it for now and we can improve later. We may as well consider this when we have the exclude filters then we can write something like:

```
+:foo-bar*
-:beta
```

Similar what TeamCity does with branch filters.

README.adoc
```diff
@@ -134,6 +134,15 @@ Next set the URL of the Gradle Enterprise instance to which the build scans shou
 
 image::enable-auto-injection.png[Enable auto-injection,{thumbnail}]
```
guylabs, line 136: I'd add the following section:

`===== VCS repository filtering (Beta)`

src/main/java/hudson/plugins/gradle/injection/InjectionUtil.java
```diff
@@ -78,4 +87,49 @@ private static boolean isEnabled(Set<String> labels, Set<String> enabledNodes) {
...
+                for (String pattern : config.getParsedInjectionVcsRepositoryPatterns()) {
+                    String url = userRemoteConfig.getUrl();
+                    if (url != null && url.contains(pattern)) {
+                        return true;
+                    }
+                }
+            }
+        }
+
+        return false;
```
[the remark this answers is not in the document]
guylabs: Agree we should skip it. If someone tampers with the data, or uninstalls the GIT repository (in that case we'd not get any data and the injection would be enabled) then we can do it as you propose.

src/main/webapp/help-injectionVcsRepositoryPatterns.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    Comma-separated list of VCS repository patterns for which to enable Gradle Enterprise Gradle plugin/Maven extension auto-injection.
+    By default, all VCS repositories have auto-injection enabled. If specified, the auto-injection will only apply to the ones which contain said patterns.
+    Disabled nodes have precedence over this filtering. This feature is currently in Beta and can be changed in future releases.
+</div>
```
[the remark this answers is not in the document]
guylabs: Also mention that it's only for GIT.

guylabs (changes requested):
I just checked the user facing parts and left some feedback.

README.adoc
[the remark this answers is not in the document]
welandaz: Exactly.

One this to note is that if we do some major rework in terms of how filter looks like or is stored, users will have to write it again, as old data is going to be erased (I'm not sure there is a way to migrate the configuration actually...and perhaps it's not worth it given the feature is in beta anyway)

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
alextu: I just meant that this test tested 2 concerns: vcs patterns and mvn configuration. My remark was just to question wether we need to have a test for those 2 concerns at the same time.

build.gradle.kts
```diff
@@ -93,6 +93,11 @@ dependencies {
         because("Lowest version that works with our dependencies")
     }
 
+    // Higher versions fail in our tests with ClassNotFoundException during SCM initialization unless Jenkins is updated
+    "optionalPluginImplementation"("org.jenkins-ci.plugins:git:4.9.4") {
```
wolfs, line 97: So with this version, you don't need to update the required Jenkins version?

welandaz: Right, this one works fine in tests

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
welandaz: As discussed, wouldn't work :(

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionGradleIntegrationTest.groovy
welandaz: As discussed, wouldn't work :(

src/main/webapp/help-injectionVcsRepositoryPatterns.html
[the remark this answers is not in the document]
welandaz: Updated

README.adoc
```diff
@@ -134,6 +134,15 @@ Next set the URL of the Gradle Enterprise instance to which the build scans shou
...
 
+It is also possible to enable injection for specific repositories by specifying VCS repository patterns. These are comma-separated values, which will be used in a contains check against the repository URL.
```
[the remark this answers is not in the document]
welandaz: Updated

README.adoc
welandaz: Added

src/main/java/hudson/plugins/gradle/injection/MavenInjectionEnvironmentContributor.java
```diff
@@ -16,25 +16,22 @@ public class MavenInjectionEnvironmentContributor extends EnvironmentContributor
...
+
+            MavenInjectionDisabledMavenOptsAction mavenInjectionDisabledMavenOptsAction = run.getAction(MavenInjectionDisabledMavenOptsAction.class);
+            if (mavenInjectionDisabledMavenOptsAction != null) {
+                envs.put(MAVEN_OPTS, mavenInjectionDisabledMavenOptsAction.mavenOpts);
+
+                // Configuration needed to support https://plugins.jenkins.io/maven-plugin/
+                envs.put(JENKINSGRADLEPLUGIN_MAVEN_PLUGIN_CONFIG_EXT_CLASSPATH, "");
+            }
+        } catch (Exception e) {
+            listener.error("Error occurred when building environment for Maven build");
```
alextu, line 34: Use logger instead + log the exception

alextu (approved):

PR #277, opened 2023-04-12
Fix the exist check of jar paths for maven-plugin

welandaz:
`GradleEnterpriseExtensionsContributorFactory` is invoked on the master node and by constructing the `FilePath` object as we do now, the `exists` check returns `false`, since those files actually are located on the agent `Node` (in case job is executed on an agent). This led to the fact that injection didn't work when maven-plugin is used and job was scheduled on a remote agent (works fine for built-in node and agents being on the same node as master)

This fix creates `FilePath` properly and the `exists` check now verifies its existence on the agent node instead of master one.
This has been tested manually using the docker setup. We would need to come up with a different test setup as well, since our current one uses either built-in node or agent being on the same filesystem.


alextu (approved):
Given this is called only by the Maven integration (I guess ?), the IO overhead is fine

c00ler (approved):

PR #282, opened 2023-04-14
Add exclusive repository filters for auto-injection configuration

welandaz:
This PR extends the work done [here](https://github.com/jenkinsci/gradle-plugin/pull/275) by changing the way filters are specified. It's now possible to specify both inclusive and exclusive filters within a single configuration block.
The format has changed from comma-separated values to newline-separated. The type of filter is distinguished using the following notation:
`+:` - specifies inclusive filter
`-:` - specifies exclusive filter


src/main/java/hudson/plugins/gradle/injection/VcsRepositoryFilter.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public static VcsRepositoryFilter of(String rawVcsRepositoryFilter) {
+        if (rawVcsRepositoryFilter == null || rawVcsRepositoryFilter.isEmpty()) {
+            return EMPTY;
+        }
+
+        List<String> inclusionFilters = new ArrayList<>();
+        List<String> exclusionFilters = new ArrayList<>();
+
+        String[] parsedFilters = rawVcsRepositoryFilter.split(SEPARATOR);
```
guylabs, line 35: Want to trim the `rawVcsRepositoryFilter` before processing it?

guylabs (approved):

src/main/java/hudson/plugins/gradle/injection/VcsRepositoryFilter.java
welandaz: I'm not sure it will work if something like this is entered:
`+:foo    \n`

README.adoc
```diff
@@ -136,10 +136,30 @@ image::enable-auto-injection.png[Enable auto-injection,{thumbnail}]
 
 ===== VCS repository filtering (Beta)
 
-It is also possible to enable injection for specific repositories by specifying VCS repository patterns. These are comma-separated values, which will be used in a contains check against the repository URL.
-For example, by specifying the filter value as `foo, bar, baz`, the injection will be performed for all Git VCS repositories which contain at least one of specified values in their VCS URL.
+It is also possible to enable/disable injection for specific repositories by specifying VCS repository filters. These are newline-delimited set of rules in the form of `+|-:repository_matching_keyword`, which will be used in a contains check against the repository URL.
```
c00ler, line 139: Maybe highlight `newline-delimited` with _italics_ or **bold**?

src/main/java/hudson/plugins/gradle/injection/GitScmListener.java
```diff
@@ -68,8 +68,8 @@ public void onCheckout(
     private static boolean isVCSInjectionFilteringNotApplied(SCM scm, InjectionConfig config) {
         return config.isDisabled()
                 || InjectionUtil.isInvalid(InjectionConfig.checkRequiredUrl(config.getServer()))
-                || config.getParsedInjectionVcsRepositoryPatterns() == null
-                || config.getParsedInjectionVcsRepositoryPatterns().isEmpty()
+                || config.getParsedVcsRepositoryFilter() == null
+                || config.getParsedVcsRepositoryFilter().isEmpty()
                 || vcsRepositoryUrlMatches(config, scm);
```
c00ler, line 73: I find method naming confusing here. `isVCSInjectionFilteringNotApplied` means no VSC filtering, but the last check does the matching (at least according to its name). Also now we can have positive and negative matching.

src/main/java/hudson/plugins/gradle/injection/InjectionConfig.java
```diff
@@ -241,21 +241,19 @@ public void setMavenInjectionDisabledNodes(List<NodeLabelItem> mavenInjectionDis
...
     @DataBoundSetter
-    public void setInjectionVcsRepositoryPatterns(String injectionVcsRepositoryPatterns) {
-        this.injectionVcsRepositoryPatterns = Util.fixEmptyAndTrim(injectionVcsRepositoryPatterns);
+    public void setVcsRepositoryFilter(String vcsRepositoryFilter) {
+        this.vcsRepositoryFilter = Util.fixEmptyAndTrim(vcsRepositoryFilter);
 
-        this.parsedInjectionVcsRepositoryPatterns = this.injectionVcsRepositoryPatterns == null
-                ? ImmutableList.of()
-                : ImmutableList.copyOf(Arrays.stream(this.injectionVcsRepositoryPatterns.split(",")).map(String::trim).collect(Collectors.toList()));
+        this.parsedVcsRepositoryFilter = VcsRepositoryFilter.of(this.vcsRepositoryFilter);
```
c00ler, line 247: Let's maybe store the raw value inside the `VcsRepositoryFilter` class. This way we will avoid storing the same data twice in the config and it will guarantee that values are always in sync.

src/main/java/hudson/plugins/gradle/injection/VcsRepositoryFilter.java
c00ler: I'm not 100% sure if it's a valid use case.

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionGradleIntegrationTest.groovy
```diff
@@ -652,23 +652,23 @@ class BuildScanInjectionGradleIntegrationTest extends BaseGradleIntegrationTest
...
-        if (pattern == "simple-") {
+        if (filter.contains("simple-")) {
             j.assertLogContains(MSG_INIT_SCRIPT_APPLIED, build2)
         } else {
             j.assertLogNotContains(MSG_INIT_SCRIPT_APPLIED, build2)
         }
 
         where:
-        pattern << ["simple-", "this-one-does-not-match"]
+        filter << ["+:simple-", "+:this-one-does-not-match"]
```
c00ler, line 668: 💄 Maybe use a table here, so that there is no data-specific logic in the test code?

```
filter                      | shouldApplyAutoInjection
"+:simple-"                 | true
"+:this-one-does-not-match" | false
```

src/main/java/hudson/plugins/gradle/injection/VcsRepositoryFilter.java
welandaz: Consider this as an input:
```
+:foo
-:bar
```

If there are additional spaces after `+:foo`, the trim of the original value won't catch them. 
Or I misunderstood your point then :)

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionGradleIntegrationTest.groovy
welandaz: Agree, will change

src/main/java/hudson/plugins/gradle/injection/GitScmListener.java
welandaz: What about calling it `shouldInjectionContinue` and also rename `vcsRepositoryUrlMatches` to `vcsRepositoryUrlInjectionEnabled`?

src/main/java/hudson/plugins/gradle/injection/InjectionConfig.java
welandaz: Agree, will change

src/main/java/hudson/plugins/gradle/injection/VcsRepositoryFilter.java
guylabs: I meant for the whole string to trim new lines at the end first before processing it.

src/main/java/hudson/plugins/gradle/injection/GitScmListener.java
c00ler: Yep, I was also thinking about splitting it.

README.adoc
welandaz: Done

src/main/java/hudson/plugins/gradle/injection/GitScmListener.java
welandaz: Done

src/main/java/hudson/plugins/gradle/injection/InjectionConfig.java
welandaz: Done

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionGradleIntegrationTest.groovy
welandaz: Done

src/main/java/hudson/plugins/gradle/injection/VcsRepositoryFilter.java
welandaz: Ah, that we already do

alextu (approved):

PR #284, opened 2023-04-18
Do not use commons-lang3 when it is not available

c00ler:
Do not use commons-lang3 as it is not available without the `maven-plugin`, that is optional.


acceptance-tests/src/test/java/hudson/plugins/gradle/AbstractAcceptanceTest.java
```diff
@@ -70,6 +70,7 @@ protected final void enableBuildScansForGradle(URI server, String agentVersion)
             settings.clickBuildScansInjection();
             settings.setGradleEnterpriseServerUrl(server);
             settings.setGradleEnterprisePluginVersion(agentVersion);
+            settings.setGradleEnterpriseAccessKey(String.format("%s=secret", server.getHost()));
```
c00ler, line 73: Setting `accessKey` triggers the execution of code paths that were not executed before. These code paths were referring to the unavailable class.

guylabs (approved):

PR #293, opened 2023-05-03
Gradle Enterprise server URL is overridable

alextu:
This PR adds a checkbox to the configuration UI to allow the GE URL configuration to override the values configured in projects. This may be useful for migrating projects from one server to another. In particular, it can be useful for projects that are already configured to upload scans to scans.gradle.com.

Same as already implemented in TC https://github.com/etiennestuder/teamcity-build-scan-plugin/pull/236

image

### Testing done
- integration test just testing presence of env var when setting the `enforceUrl` config flag
- acceptance test setting the `enforceUrl` flag, building a project with a preconfigured GE URL publishes to the global configured GE URL. Note that it does not test that the `allowUntrustedServer` is also overridden. This has been tested manually.

### Submitter checklist


README.adoc
```diff
@@ -134,6 +134,8 @@ Next set the URL of the Gradle Enterprise instance to which the build scans shou
 
 image::enable-auto-injection.png[Enable auto-injection,{thumbnail}]
```
c00ler, line 135: Do we need to update the screenshot to the new one that includes the checkbox?

acceptance-tests/src/main/java/hudson/plugin/gradle/BuildScansInjectionSettings.java
```diff
@@ -43,6 +43,18 @@ public void setGradleEnterpriseServerUrl(URI server) {
...
+    public void clickEnforceUrl() {
+        ensureConfigPage();
+
+        control(by.checkbox("Enforce Gradle Enterprise server url")).click();
+    }
+
+    public void clickAllowUntrustedServer() {
+        ensureConfigPage();
+
+        control(by.checkbox("Allow untrusted server")).click();
```
c00ler, line 55: We use these 2 methods together quite often now, worth extracting the method. 

No need to do it in this PR can be done later.

acceptance-tests/src/test/java/hudson/plugins/gradle/GradleInjectionTest.java
```diff
@@ -275,6 +275,36 @@ public void injectionWorksOnAgents() throws Exception {
         assertBuildScanPublished(build);
     }
 
+    @Test
+    public void setEnforceUrlPublishesBuildScan() throws Exception {
+        // given
+        setEnforceUrl();
+        setAllowUntrustedServer();
+        String expectedAgentVersion = "3.10.3";
```
c00ler, line 283: Can we use the latest version here?

src/main/resources/hudson/plugins/gradle/injection/InjectionConfig/help-enforceUrl.html
```diff
@@ -0,0 +1,3 @@
+<div>
+    Whether to enforce the Gradle Enterprise Server URL configured in this connection over a URL configured in the project's build.
```
c00ler, line 2: It must be phrased differently I think. There is no such thing as a connection. Connection is a TC-specific thing.

acceptance-tests/src/main/java/hudson/plugin/gradle/BuildScansInjectionSettings.java
alextu: done in this PR

c00ler (approved):

PR #294, opened 2023-05-04
Add Gradle Enterprise exceptions log filter

c00ler:
Search for Gradle Enterprise agent errors and show a warning if errors are detected:

Screenshot 2023-05-06 at 14 44 23

The feature requires an explicit opt-in from the user:

Screenshot 2023-05-06 at 14 55 53

**TODO**:

- [ ] ~update README~ - not needed, since this functionality is not yet available for the user
- [x] test with new agents when released


src/main/java/hudson/plugins/gradle/GradleLogProcessor.java
welandaz, line 1: I still don't see the value for having this interface. It's only used via `AbstractGradleLogProcessor` and there is currently no use for implementing it separately. I'd move `processLogLine` to `AbstractGradleLogProcessor` as abstract method. 

Do you have some more plans for it? Or perhaps forsee the usage of it elsewhere?

src/test/groovy/hudson/plugins/gradle/injection/GradleEnterpriseExceptionLogProcessorTest.groovy
```diff
@@ -0,0 +1,78 @@
...
+            scanDetails.isEmpty()
+            !hasMavenErrors
+            hasGradleErrors
+        }
+    }
+
+    def "detects errors for Maven and Gradle"() {
+        when:
+        processor.processLogLine("Internal error in Gradle Enterprise Gradle plugin: com.acme.FooBar")
+        processor.processLogLine("[ERROR] Internal error in Gradle Enterprise Maven extension: com.acme.FooBar")
```
welandaz, line 54: How about extracting these into constants? They are used in other tests too

welandaz (approved):
Great work!
Left some small comments

src/main/java/hudson/plugins/gradle/GradleLogProcessor.java
c00ler: Agree. Let's keep it simple. Removed.

c00ler:
@welandaz @alextu For now I've removed the toggle from the UI: https://github.com/jenkinsci/gradle-plugin/pull/294/commits/7a8f660cea860970ede2a3719f77bbfb2c6c89c0 The feature will not be available for the users.

welandaz (approved):

PR #300, opened 2023-05-26
Add page to gather Gradle Enterprise unhandled errors

alextu:
Follow up of https://github.com/jenkinsci/gradle-plugin/pull/294

Once the administrator enables "Check for the Gradle Enterprise build agent errors", a new top level "Gradle Enterprise" button shows up that redirects to a page showing all builds having Gradle Enterprise unhandled errors. Both the button and the page are restricted to admins.

image

image

### Testing done
- acceptance test capturing an error on a Freestyle build and ensuring it's there in this new page
- integration test ensuring capturing errors from logs works as well for Pipeline builds

### Submitter checklist


acceptance-tests/src/test/java/hudson/plugins/gradle/GradleEnterpriseErrorsTest.java
```diff
@@ -0,0 +1,95 @@
...
+import org.openqa.selenium.By;
+import org.openqa.selenium.WebElement;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.not;
+import static org.jenkinsci.test.acceptance.Matchers.containsString;
+
+public class GradleEnterpriseErrorsTest extends AbstractAcceptanceTest {
+
+    private static final String AGENT_VERSION = "3.13.2";
```
welandaz, line 17: Just a nitpick - 3.13.3 is the latest :)

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionGradleIntegrationTest.groovy
```diff
@@ -111,6 +111,63 @@ class BuildScanInjectionGradleIntegrationTest extends BaseGradleIntegrationTest
...
+
+        when:
+        // first build to download Gradle
+        def build = j.buildAndAssertSuccess(pipelineJob)
+
+        then:
+        j.assertLogNotContains(MSG_INIT_SCRIPT_APPLIED, build)
+
+        when:
+        // first build to download Gradle
```
welandaz, line 150: This is now the second run, right?

src/main/java/hudson/plugins/gradle/injection/GradleEnterpriseErrorsAction.java
```diff
@@ -0,0 +1,108 @@
...
+        return RunList.fromJobs((Iterable) Jenkins.get().allItems(Job.class))
+            .limit(200)
+            .completedOnly()
+            .stream()
+            .filter(r -> {
+                AbstractBuildScanAction action = ((Run) r).getAction(AbstractBuildScanAction.class);
+                return action != null && (action.getHasGradleErrors() || action.getHasMavenErrors());
+            })
+            .map(r -> GeErrorModel.fromRun((Run) r))
+            .iterator();
```
welandaz, line 49: Is the idea to show at most 200 errors? In that case I'd filter first and then limit later, as now out of 200 taken, there might be no errors at all even though there were some

src/main/java/hudson/plugins/gradle/injection/GradleEnterpriseErrorsAction.java
```diff
@@ -0,0 +1,108 @@
...
+        return InjectionConfig.get().isCheckForBuildAgentErrors();
+    }
+
+    public static class GeErrorModel {
+        private final String buildTool;
+        private final String project;
+        private final String buildStatusIconClassName;
+        private final String buildStatus;
+        private final String buildUrl;
+        private final Date start;
```
welandaz, line 62: Do we need to use `Date` because of Jenkins limitations? Or some other reason to not opt for `java.time`?

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionGradleIntegrationTest.groovy
```diff
@@ -111,6 +111,63 @@ class BuildScanInjectionGradleIntegrationTest extends BaseGradleIntegrationTest
...
+        def firstRun = j.buildAndAssertSuccess(pipelineJob)
+
+        then:
+        j.assertLogNotContains(MSG_INIT_SCRIPT_APPLIED, firstRun)
+
+        when:
+        enableBuildInjection(agent, gradleVersion)
+        withInjectionConfig {
+            checkForBuildAgentErrors = true
+        }
```
welandaz, line 160: I think you should either swap lines `157` and `158` places, or restart a slave agent to propagate the changes. The `withInjectionConfig` doesn't restart it, so I don't this the change was applied. Does the assertion below really work for you?

welandaz (commented):
Nice work! 👍 
Left a few minor comments

src/main/java/hudson/plugins/gradle/injection/GradleEnterpriseErrorsAction.java
alextu: Yeah I checked the UI component showing dates, it only supports `Date` 😞

src/main/java/hudson/plugins/gradle/injection/GradleEnterpriseErrorsAction.java
alextu: Good catch 👍 !

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionGradleIntegrationTest.groovy
alextu: 🤔 Why a restart would be required ? This flag is only read in memory on master, it does not need propagation to agents (and even if it was, we should not need to restart thanks `InjectionConfigChangeListener`)

etiennestuder:
Some comments on the UI:
- is it possible to center the values for the _build tool_ and the _build outcome_?
- are the icons used for the build outcome default Jenkins icons or our own? if we picked the icons, I suggest we use the same icons we use in our build scan app for build failed/succeeded.
- I would rename the title from _Unhandled errors_ to _Unexpected errors_
- can we add some description under _Unexpected errors_ ? For example:

**Unexpected errors**
Errors that were not gracefully handled by the Gradle Enterprise plugin

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionGradleIntegrationTest.groovy
welandaz: So I ran this test locally and this last assertion fails for me. 
The `InjectionConfigChangeListener` would definitely do the job (I've swapped the lines and it didn't help either, so I'm guessing the problem is somewhere else)

src/main/java/hudson/plugins/gradle/injection/GradleEnterpriseErrorsAction.java
```diff
@@ -0,0 +1,108 @@
...
+            this.project = project;
+            this.buildStatusIconClassName = buildStatusIconClassName;
+            this.buildStatus = buildStatus;
+            this.buildUrl = buildUrl;
+            this.start = start;
+        }
+
+        static GeErrorModel fromRun(Run<?, ?> r) {
+            AbstractBuildScanAction action = r.getAction(AbstractBuildScanAction.class);
+            return new GeErrorModel(action.getHasGradleErrors() ? "GRADLE" : "MAVEN",
```
c00ler, line 76: We will be adding more build tools. It's better not to fall back to Gradle by default. 

Related note. At the moment the logic is split between 2 methods, `filter`, which checks for the error presence, and this method, which is assumed to be called only if the error exists. I think we can simplify it and do the check only here and return `Optional` (if no error or unsupported build tool) and have a filter just based on the returned `Optional`. WDYT?

src/main/resources/hudson/plugins/gradle/injection/GradleEnterpriseErrorsAction/index.jelly
```diff
@@ -0,0 +1,51 @@
...
+                            <th class="small-column" initialSortDir="up">${%Start time}</th>
+                            <th class="medium-column">${%URL}</th>
+                        </tr>
+                    </thead>
+                    <j:forEach items="${it.errors}" var="e">
+                        <tr>
+                            <td class="medium-column">${e.project}</td>
+                            <td class="small-column center">
+                                <j:choose>
+                                    <j:when test="${e.buildTool == 'GRADLE'}">
```
c00ler, line 28: Since all the logic is on the server side, maybe we can just return an icon from the server side as well? Then we don't need any conditions here and all the related logic will be in one place. It will be easier to add new tools in the future.

c00ler (changes requested):

src/main/resources/hudson/plugins/gradle/injection/GradleEnterpriseErrorsAction/index.jelly
```diff
@@ -0,0 +1,51 @@
...
+    <l:layout title="Gradle Enterprise" permission="${app.ADMINISTER}">
+        <l:main-panel>
+            <div class="jenkins-app-bar">
+                <div class="jenkins-app-bar__content">
+                    <h1>Gradle Enterprise</h1>
+                </div>
+            </div>
+            <section class="jenkins-section">
+                <h2 class="jenkins-section__header">${%Unexpected errors}</h2>
+                <p>Errors that were not gracefully handled by the Gradle Enterprise plugin</p>
```
c00ler, line 12: Maybe a bit more verbose here?

```suggestion
                <p>Errors that were not gracefully handled by the Gradle Enterprise Gradle plugin/Gradle Enterprise Maven extension</p>
```

src/main/java/hudson/plugins/gradle/injection/GradleEnterpriseErrorsAction.java
alextu: Yes, that sounds better, also I just realized that we need to support potentially multiple build tools errors per run, right ?

src/main/resources/hudson/plugins/gradle/injection/GradleEnterpriseErrorsAction/index.jelly
alextu: yep, thought about this as well, will refactor towards this

alextu:
Latest changes:
- Bump minimal Jenkins version to LTS `2.303.3`
- Adapt styles to the updated Jenkins (see below)
- Handle multiple build agent errors
- Update chrome driver
- Code review fixes
- Spotbug fixes

image

c00ler (approved):

PR #310, opened 2023-06-18
Build on container agents

NotMyFault:
Container agents are usually faster to provision than a VM.


alextu (commented):
I don't see `useContainerAgent` taken into account in `https://github.com/jenkins-infra/pipeline-library/blob/master/vars/buildPluginWithGradle.groovy` 🤔

NotMyFault:
May not work with dockerized and docker-based tests too.

PR #312, opened 2023-06-22
Fix credentials not masked when `DurableTaskStep.USE_WATCHING` is true

alextu:
This should fix https://issues.jenkins.io/browse/JENKINS-71509

Although this fixes the credentials masking when `org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep.USE_WATCHING` is set to true, the original `GradleEnterpriseExceptionTaskListenerDecorator` does not work in this case. It does not "see" all the log lines, it will require more investigation.

### Testing done
Provided integration tests
Manual tested on a local instance with the OpenTelemetry plugin (it sets `org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep.USE_WATCHING` to true) that secrets are masked on:
- Simple pipeline build with credentials binding plugin
- Simple pipeline build with Hashicorp Vault plugin
- Simple freestyle build with Hashicorp Vault plugin

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/hudson/plugins/gradle/injection/GradleEnterpriseExceptionTaskListenerDecoratorFactory.java
+++ b/src/main/java/hudson/plugins/gradle/injection/GradleEnterpriseExceptionTaskListenerDecoratorFactory.java
@@ -1,34 +1,43 @@
 package hudson.plugins.gradle.injection;
 
 import hudson.Extension;
+import hudson.model.Queue;
 import hudson.model.Run;
+import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.jenkinsci.plugins.workflow.log.TaskListenerDecorator;
-import org.jenkinsci.plugins.workflow.steps.DynamicContext;
 
-import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.Serializable;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 @SuppressWarnings("unused")
 @Extension
-public class GradleEnterpriseExceptionTaskListenerDecoratorFactory extends DynamicContext.Typed<TaskListenerDecorator> {
+public class GradleEnterpriseExceptionTaskListenerDecoratorFactory implements TaskListenerDecorator.Factory {
 
-    @Nonnull
-    @Override
-    protected Class<TaskListenerDecorator> type() {
-        return TaskListenerDecorator.class;
-    }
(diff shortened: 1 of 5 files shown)


src/main/java/hudson/plugins/gradle/injection/GradleEnterpriseExceptionTaskListenerDecoratorFactory.java
```diff
@@ -1,34 +1,43 @@
...
+public class GradleEnterpriseExceptionTaskListenerDecoratorFactory implements TaskListenerDecorator.Factory {
 
-    @Nonnull
-    @Override
-    protected Class<TaskListenerDecorator> type() {
-        return TaskListenerDecorator.class;
-    }
+    private static final Logger LOGGER = Logger.getLogger(GradleEnterpriseExceptionTaskListenerDecoratorFactory.class.getName());
 
-    @CheckForNull
```
c00ler, line 23: ❓ Why it was removed? The method returns `null`.

src/test/groovy/hudson/plugins/gradle/BaseGradleIntegrationTest.groovy
```diff
@@ -9,6 +16,7 @@ import org.junit.rules.RuleChain
 abstract class BaseGradleIntegrationTest extends AbstractIntegrationTest {
 
     public final GradleInstallationRule gradleInstallationRule = new GradleInstallationRule(j)
+    static final String GRADLE_ENTERPRISE_PLUGIN_VERSION = '3.13.2'
```
c00ler, line 19:
```suggestion
    static final String GRADLE_ENTERPRISE_PLUGIN_VERSION = '3.13.4'
```

src/test/groovy/hudson/plugins/gradle/injection/GradleSnippets.groovy
```diff
@@ -0,0 +1,51 @@
+package hudson.plugins.gradle.injection
+
+import hudson.plugins.gradle.GradleInstallationRule
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition
+import org.jenkinsci.plugins.workflow.job.WorkflowJob
+import org.jvnet.hudson.test.JenkinsRule
+
+class GradleSnippets {
+
+    def static pipelineJobWithError(JenkinsRule j, GradleInstallationRule gradleInstallationRule) {
```
c00ler, line 10: I think it's better to use a concrete type instead of `def`. It's not easy to understand what's being returned here

c00ler (approved):

src/main/java/hudson/plugins/gradle/injection/GradleEnterpriseExceptionTaskListenerDecoratorFactory.java
alextu: Just got lost during refactoring, thanks

PR #369, opened 2023-11-21
Use Chrome for testing

alextu:
- Configure the Chrome driver to use the Chrome for testing binary.
- Also do not run acceptance tests on jenkins.io where neither firefox, chrome, firefox-container and chrome-container are currently working for us. 
In particular because of https://github.com/jenkinsci/acceptance-test-harness/issues/1170 on firefox-container, but we still have failing acceptance tests on jenkins.io with chrome-container, that do not happen locally or on other infra.

### Testing done

Ran build on jenkins.io and internally.

```[tasklist]
### Submitter checklist
```


acceptance-tests/src/main/java/hudson/plugin/gradle/ath/config/WebDriverProvider.java
```diff
@@ -0,0 +1,97 @@
...
+        try {
+            return fallbackConfig.createWebDriver(testCleaner, testName, time);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private WebDriver createChromeWebDriver() {
+        ChromeOptions chromeOptions = new ChromeOptions();
+        chromeOptions.setBinary("/usr/bin/google-chrome-for-testing");
```
zielezin, line 65: Using `/usr/bin/google-chrome-for-testing` would point you to the latest version and as soon as we would update this would conflict with your chromedriver version.

I would change this to: `/opt/google/119.0.6045.105/chrome-linux64/chrome`

```
zielezin@dev34:~$ /opt/google/119.0.6045.105/chrome-linux64/chrome --version
Google Chrome for Testing 119.0.6045.105
```

zielezin (approved):
LGTM.

I have no experience with running this setup so I assume you have tested this ;)

welandaz (approved):

acceptance-tests/src/main/java/hudson/plugin/gradle/ath/config/WebDriverProvider.java
alextu: So actually, after investigating, it turned out we can delegate all the driver plumbing to Selenium. It will try to find the specified Chrome version installed and if not found, download it. So my latest commit specifies the version to `stable` which should afaik try to find a stable Chrome installed, and if not present will download the latest stable.

acceptance-tests/build.gradle.kts
```diff
@@ -5,7 +5,6 @@ import java.net.URL
 
 plugins {
     java
-    id("buildlogic.chromedriver")
```
zielezin, line 8: As Selenium does this I think this plugin can be deleted too.,

alextu: Yes it's deleted there https://github.com/jenkinsci/gradle-plugin/pull/369/files#diff-00d1422a52d1cd37c7e2205657f1533c8db46efe0adcac04dc60cd157a6addda

PR #374, opened 2023-12-11
Fix GradleConsoleAnnotator when Timestamper is globally enabled

alextu:
Fixes [JENKINS-72411](https://issues.jenkins.io/browse/JENKINS-72411) by detecting the timestamp prefix and shift the Gradle notes.
It should always be a complete timestamp and its format should not change (afaik it's changed by a formatter later on when displaying the console).

### Testing done
- Unit test for detecting the timestamp prefix
- Integration test to check the expected Gradle notes. The left menu could not be part of this test as it's loaded separately with an ajax request. I checked it manually though.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/hudson/plugins/gradle/GradleConsoleAnnotator.java
+++ b/src/main/java/hudson/plugins/gradle/GradleConsoleAnnotator.java
@@ -4,6 +4,9 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.charset.Charset;
 
+import static hudson.plugins.gradle.TimestampPrefixDetector.detectTimestampPrefix;
+import static hudson.plugins.gradle.TimestampPrefixDetector.trimTimestampPrefix;
+
 /**
  * @author ikikko
  * @see <a href="https://github.com/jenkinsci/ant-plugin/blob/master/src/main/java/hudson/tasks/_ant/AntConsoleAnnotator.java">AntConsoleAnnotator</a>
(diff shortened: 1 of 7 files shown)


src/main/java/hudson/plugins/gradle/TimestampPrefixDetector.java
```diff
@@ -0,0 +1,24 @@
...
+    static int detectTimestampPrefix(String line) {
+        Matcher matcher = TimestampPatternR.matcher(line);
+        if (matcher.matches()) {
+            return matcher.group(1).length();
+        }
+        return 0;
+    }
+
+
+}
```
welandaz, line 24: I'd make the class final and add private constructor

src/main/java/hudson/plugins/gradle/GradleConsoleAnnotator.java
```diff
@@ -28,7 +33,12 @@ protected void processLogLine(String line) throws IOException {
         line = trimEOL(line);
 
         if (annotateGradleOutput) {
-            if (line.startsWith(":") || line.startsWith("> Task :")) { // put the annotation
+            // assumption that the timestamp prefix will be present or not for all lines
+            if (timestampPrefix == null) {
+                timestampPrefix = detectTimestampPrefix(line);
+            }
```
welandaz, line 39: Can it be the case that one line has the timestamp and the next one doesn't?

alextu: From what I saw, it should not be the case. I think it's ok to take this risk/assumption vs detecting the timestamp on every log line.

welandaz (approved):

PR #393, opened 2024-02-08
Apply Develocity Maven extension only if not already applied

welandaz:
If a project already has Develocity Maven extension applied, we shouldn't override it with our injection.

If Develocity Maven extension injection is enabled, we will check the `.mvn/extensions.xml` after checkout for the presence of said extension and disable the injection.

### DIFF
--- a/src/main/java/hudson/plugins/gradle/injection/GitScmListener.java
+++ b/src/main/java/hudson/plugins/gradle/injection/GitScmListener.java
@@ -20,6 +20,7 @@ import org.slf4j.LoggerFactory;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import java.io.File;
+import java.io.IOException;
 import java.util.List;
 
 import static hudson.plugins.gradle.injection.GradleInjectionAware.JENKINSGRADLEPLUGIN_GRADLE_ENTERPRISE_GRADLE_INJECTION_ENABLED;
@@ -31,6 +32,8 @@ import static hudson.plugins.gradle.injection.MavenOptsHandler.MAVEN_OPTS;
 public class GitScmListener extends SCMListener {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(GitScmListener.class);
+
+    private static final MavenCoordinates DEVELOCITY_EXTENSION_MAVEN_COORDINATES = new MavenCoordinates("com.gradle", "gradle-enterprise-maven-extension");
 
     @Override
     public void onCheckout(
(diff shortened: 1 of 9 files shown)


src/main/java/hudson/plugins/gradle/injection/MavenExtensions.java
```diff
@@ -0,0 +1,71 @@
...
+    private MavenExtensions(Document document) {
+        this.document = document;
+    }
+
+    static MavenExtensions empty() {
+        return new MavenExtensions(null);
+    }
+
+    static MavenExtensions fromFilePath(FilePath extensionsFile) {
+        try(InputStream inputStream = extensionsFile.read()) {
```
welandaz, line 42: Closing the InputStream here as it's not clear whether DocumentBuilder.parse() will do it

src/main/java/hudson/plugins/gradle/injection/GitScmListener.java
```diff
@@ -121,6 +128,25 @@ private static boolean shouldDisableMavenInjection(InjectionConfig config) {
         return config.isInjectMavenExtension();
     }
 
+    private static boolean mavenExtensionAlreadyApplied(InjectionConfig config, FilePath workspace) throws IOException, InterruptedException {
+        if (!config.isInjectMavenExtension()) {
+            return false;
+        }
```
welandaz, line 134: If injection of Maven extension is tuned off, we don't need to perform this check and simply leave things as they are

src/main/java/hudson/plugins/gradle/injection/MavenExtensions.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @SuppressWarnings("BooleanMethodIsAlwaysInverted")
+    boolean hasExtension(MavenCoordinates coordinates) {
+        if (document == null || coordinates == null) {
+            return false;
+        }
+
+        String expr = String.format(EXTENSION_XPATH_EXPR, coordinates.groupId(), coordinates.artifactId());
+        try {
+            XPathExpression exprCompiled = XPATH.compile(expr);
```
alextu, line 61: I thought we would pay a small price to compile it every time, but it looks like they already have some internal optimization: first compile call takes about ~6ms, next one 0ms

alextu (approved):
Good job!
Some notes:
- I've tested it manually with a docker agent, works fine for both freestyle and pipeline
- This is kind of breaking change, ie if users had the injection configured on projects, it won't publish anymore to the globally configured DV server. I assume it's ok if it's properly communicated in release notes?

src/main/java/hudson/plugins/gradle/injection/MavenExtensions.java
welandaz: The method expects any Maven Coordinates, but in this case we can rework it just for Develocity Maven Extension (so we can precompile).
Do you think it's worth it?

welandaz:
> Good job! Some notes:
> 
> * I've tested it manually with a docker agent, works fine for both freestyle and pipeline
> * This is kind of breaking change, ie if users had the injection configured on projects, it won't publish anymore to the globally configured DV server. I assume it's ok if it's properly communicated in release notes?

Right, that would be the case for Maven ones. The "enforce url" feature works when using Gradle plugin from what I can see, so there would be no way to override it for Maven extension use case.

alextu:
> > Good job! Some notes:
> > 
> > * I've tested it manually with a docker agent, works fine for both freestyle and pipeline
> > * This is kind of breaking change, ie if users had the injection configured on projects, it won't publish anymore to the globally configured DV server. I assume it's ok if it's properly communicated in release notes?
> 
> Right, that would be the case for Maven ones. The "enforce url" feature only works when using Gradle plugin from what I can see, so there would be no way to override it for Maven extension use case.
> 
> We can introduce it though in a separate PR, probably makes sense even for the next release to mitigate this somehow

That's a good point, I think it makes sense to release both, so when the enforce url is true, the extension is taken from the project but the URL is overridden.

PR #395, opened 2024-02-08
Respect enforce URL configuration option for Maven builds

welandaz:
The "Enforce Develocity server url" configuration option was only applicable for Gradle builds. This PR adds support for it for Maven ones.

### DIFF
--- a/src/main/java/hudson/plugins/gradle/injection/MavenOptsHandler.java
+++ b/src/main/java/hudson/plugins/gradle/injection/MavenOptsHandler.java
@@ -41,7 +41,7 @@ final class MavenOptsHandler {
                 .map(SystemProperty::asString)
                 .collect(Collectors.joining(SPACE));
 
-        return filtered(getCurrentMavenOpts(node))
+        return filtered(getCurrentMavenOpts(node), Collections.emptySet())
                 .map(current -> String.join(SPACE, current, additionalProperties))
                 .orElse(additionalProperties);
     }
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/gradle/injection/MavenOptsHandler.java
```diff
@@ -102,7 +106,7 @@ private String filterMavenOpts(String mavenOpts) {
         return filtered;
     }
 
-    private boolean shouldBeKept(String systemProperty) {
-        return keys.stream().noneMatch(systemProperty::contains);
+    private boolean shouldBeKept(String systemProperty, Set<String> keepKeys) {
+        return keys.stream().filter(it -> !keepKeys.contains(it)).noneMatch(systemProperty::contains);
```
alextu, line 110: My brain hangs when trying to figure what this does 🥲 , I guess `Sets.difference(keys, keepKeys).stream().noneMatch(systemProperty::contains);` is a bit more expressive but let's keep it that way.

alextu (approved):

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
```diff
@@ -731,17 +735,28 @@ node {
...
+            j.assertLogContains("Publishing build scan...", build)
+        } else {
+            // Project has localhost:8080 configured
+            j.assertLogContains("[WARNING] No build scan will be published: Gradle Enterprise features were not enabled due to an unexpected error while contacting Gradle Enterprise", build)
+        }
 
         where:
-        mavenSetup << ["withEnv([\"PATH+MAVEN=\${tool 'mavenInstallationName'}/bin\"]) {" , "withMaven(maven: 'mavenInstallationName') {"]
+        mavenSetup                                                          | isUrlEnforced
+        "withEnv([\"PATH+MAVEN=\${tool 'mavenInstallationName'}/bin\"]) {"  | false
```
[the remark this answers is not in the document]
welandaz: it's testing 2 different ways maven can be configured in the pipeline (like with maven plugin for example)
We do injection differently for the plugin case, so I have this for sanity check purposes mostly

src/main/java/hudson/plugins/gradle/injection/MavenOptsHandler.java
welandaz: Yeah, `Sets.difference` is better, will replace

welandaz: Done

PR #398, opened 2024-02-22
Disable injection when custom extensions are detected

alextu:
Follow up of https://github.com/jenkinsci/gradle-plugin/pull/393

This PR adds 2 new fields to the Develocity Maven injection:

image

Those are used in case built projects have defined custom Develocity derived extensions:
- If the user specifies a GAV in "Develocity Maven Extension Custom Coordinates" field, the Develocity injection will be disabled if a corresponding extension is found in `.mvn/extensions` of the currently built project.
- Similarly, if the user specifies a GAV in "Common Custom User Data Maven Extension Custom Coordinates" field, the CCUD injection will be disabled if a corresponding extension is found in `.mvn/extensions` of the currently built project.

It's a bit complex in terms of implementation since we need to parse the existing `MAVEN_OPTS` looking for the maven ext classpath system property, and filter it accordingly.
We might want to revisit this since it looks like the `ScmListener` we use is called in many cases, even when no Scm is configured...

### Testing done

Unit tests and a few integration tests. Some cases cannot easily be setup as integration tests since they would require "real" custom Develocity extensions.
I also tested manually the 2 cases:
- custom extension defined in the built project and corresponding "Develocity Maven Extension Custom Coordinates" field set => DV extension is not applied. The CCUD extension is still applied if "Enable Common Custom User Data Maven extension auto-injection" is checked.
- custom extension defined in the built project and corresponding "Common Custom User Data Maven Extension Custom Coordinates" field set => CCUD extension is not applied. Note that the DV extension is still applied, although most likely, it should also be defined locally; in that case, it's not applied.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/hudson/plugins/gradle/injection/GitScmListener.java
+++ b/src/main/java/hudson/plugins/gradle/injection/GitScmListener.java
@@ -1,7 +1,6 @@
 package hudson.plugins.gradle.injection;
 
 import com.google.common.base.Strings;
-import com.google.common.collect.Sets;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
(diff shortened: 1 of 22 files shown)


welandaz (approved):
It is a bit convoluted now, but in order to simplify it, we need to rewrite the way injection is done in general.
Good work otherwise 👍

PR #400, opened 2024-02-28
Add an ability to capture task/goal input files

welandaz:
This PR adds an ability to enable/disable capturing goal/task input files in a build scan (more information can be found [here](https://docs.gradle.com/enterprise/maven-extension/#capturing_goal_input_files) and [here](https://docs.gradle.com/enterprise/gradle-plugin/#capturing_task_input_files))

For DV Maven Extension, we configure this via a system property. For DV Gradle Plugin, the init script was adjusted accordingly to support old and new versions of Gradle and the plugin itself.

In UI configuration, the toggle is called `Capture Task Input Files` in Gradle Settings and `Capture Goal Input Files` in Maven Settings

gradle-settings
maven-settings


src/main/resources/hudson/plugins/gradle/injection/InjectionConfig/config.jelly
```diff
@@ -98,6 +98,9 @@
                         </f:entry>
                     </f:repeatableProperty>
                 </f:entry>
+                <f:entry field="gradleCaptureTaskInputFiles" help="/plugin/gradle/help-captureFileInputs.html">
```
alextu, line 101: I think you can avoid adding the `help` attribute and just name by convention the tooltip file `help-gradleCaptureTaskInputFiles.html`

src/main/resources/hudson/plugins/gradle/injection/InjectionConfig/config.jelly
```diff
@@ -133,6 +136,9 @@
                         </f:entry>
                     </f:repeatableProperty>
                 </f:entry>
+                <f:entry field="mavenCaptureGoalInputFiles" help="/plugin/gradle/help-captureFileInputs.html">
```
alextu, line 139: Same here

alextu (approved):

src/main/resources/hudson/plugins/gradle/injection/InjectionConfig/config.jelly
welandaz: Yup, works, had to move them to a different folder though

PR #408, opened 2024-03-12
Fix task input files capturing for Gradle 5.x in afterevaluate

welandaz:
This PR fixes the task input files capturing for Gradle 5.x in afterevaluate block when enforce URL option is chosen (we simply move it out of that block)

### DIFF
--- a/src/main/resources/hudson/plugins/gradle/injection/init-script.gradle
+++ b/src/main/resources/hudson/plugins/gradle/injection/init-script.gradle
@@ -36,6 +36,7 @@ initscript {
             maven {
                 url pluginRepositoryUrl
                 if (gradlePluginRepoUsername && gradlePluginRepoPassword) {
+                    logger.lifecycle("Using credentials for plugin repository")
                     credentials {
                         username(gradlePluginRepoUsername)
                         password(gradlePluginRepoPassword)
@@ -95,6 +96,7 @@ def geEnforceUrl = Boolean.parseBoolean(getInputParam('jenkinsGradlePlugin.gradl
 def gePluginVersion = getInputParam('jenkinsGradlePlugin.gradle-enterprise.plugin.version')
 def ccudPluginVersion = getInputParam('jenkinsGradlePlugin.ccud.plugin.version')
 
+def atLeastGradle5 = GradleVersion.current() >= GradleVersion.version('5.0')
 def atLeastGradle4 = GradleVersion.current() >= GradleVersion.version('4.0')
 
 // finish early if configuration parameters passed in via system properties are not valid/supported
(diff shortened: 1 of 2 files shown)


src/main/resources/hudson/plugins/gradle/injection/init-script.gradle
```diff
@@ -137,12 +139,12 @@ if (GradleVersion.current() < GradleVersion.version('6.0')) {
...
                             buildScan.server = geUrl
                             buildScan.allowUntrustedServer = geAllowUntrustedServer
-                            if (isAtLeast(gePluginVersion, '2.1')) {
-                                if (isNotAtLeast(gePluginVersion, '3.7')) {
-                                    buildScan.captureTaskInputFiles = geCaptureTaskInputFiles
-                                } else {
-                                    buildScan.capture.taskInputFiles = geCaptureTaskInputFiles
-                                }
+                        }
+                        if (geCaptureTaskInputFiles && isAtLeast(gePluginVersion, '2.1') && atLeastGradle5) {
```
alextu, line 143: Is this block executed if https://github.com/jenkinsci/gradle-plugin/pull/408/files#diff-9026492162e503f2dfa14b8cb79801da9f4fd7425809e813bb2413b4f728bc0bL116-L132 is ? ie do we need to have it duplicated if we want to set it in all cases?

src/main/resources/hudson/plugins/gradle/injection/init-script.gradle
```diff
@@ -188,11 +190,11 @@ if (GradleVersion.current() < GradleVersion.version('6.0')) {
                     logger.lifecycle("Enforcing Develocity: $geUrl, allowUntrustedServer: $geAllowUntrustedServer, captureTaskInputFiles: $geCaptureTaskInputFiles")
                     ext.server = geUrl
                     ext.allowUntrustedServer = geAllowUntrustedServer
-                    if (isAtLeast(gePluginVersion, '2.1')) {
-                        if (isNotAtLeast(gePluginVersion, '3.7')) {
-                            ext.buildScan.captureTaskInputFiles = geCaptureTaskInputFiles
-                        } else {
+                    if (geCaptureTaskInputFiles && isAtLeast(gePluginVersion, '2.1')) {
```
alextu, line 193: Same question here

alextu (commented):
Added a comment regarding what parts of the init script are executed when the plugin is already set or not

src/main/resources/hudson/plugins/gradle/injection/init-script.gradle
welandaz: Not necessarily; one block is executed if the plugin is not already applied and the other one is when enforceUrl is required (which happens for both cases where plugin is or isn't already applied).

welandaz: The logic is now extracted

alextu (approved):

PR #409, opened 2024-03-14
Update Develocity injection for new Develocity plugin artifacts

leonard84:
This PR updates the injection of Develocity plugins and supports configuring both old Gradle Enterprise and new Develocity plugins.

This is currently blocked until the final versions of the Develocity plugin/extension has been released.

### Testing done

Tested with existing integration tests, as they cover the changed code.

```[tasklist]
### Submitter checklist
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Link to relevant issues in GitHub or Jira
```


src/main/java/hudson/plugins/gradle/injection/InitScriptTemplate.java
```diff
@@ -0,0 +1,41 @@
...
+
+    static {
+        Map<String, String> replacements = new HashMap<>();
+        Arrays.stream(InitScriptVariables.values())
+            .filter(v -> !v.getTemplateName().isEmpty())
+            .forEach(v -> replacements.put(v.getTemplateName(), v.getTemplateTargetName()));
+        REPLACEMENTS = Collections.unmodifiableMap(replacements);
+    }
+
+    public static String replacePlaceholders(String template, Map<String, String> replacements) {
```
alextu, line 22: 🤔 Should we avoid this replacement and rename the env vars to remove the `JENKINSGRADLEPLUGIN` prefix?

src/main/java/hudson/plugins/gradle/injection/MavenExtensionsDetector.java
```diff
@@ -25,7 +25,10 @@ static Set<MavenExtension> detect(InjectionConfig config, FilePath workspace) th
             MavenExtensions mavenExtensions = MavenExtensions.fromFilePath(extensionsFile);
 
             Set<MavenExtension> knownExtensions = new HashSet<>();
-            if (mavenExtensions.hasExtension(MavenExtension.GRADLE_ENTERPRISE.getCoordinates()) ||
+            if (mavenExtensions.hasExtension(MavenExtension.DEVELOCITY.getCoordinates()) ||
+                mavenExtensions.hasExtension(MavenCoordinates.parseCoordinates(config.getMavenExtensionCustomCoordinates()))) {
+                knownExtensions.add(MavenExtension.DEVELOCITY);
+            } else if (mavenExtensions.hasExtension(MavenExtension.GRADLE_ENTERPRISE.getCoordinates()) ||
                 mavenExtensions.hasExtension(MavenCoordinates.parseCoordinates(config.getMavenExtensionCustomCoordinates()))) {
```
alextu, line 32: 🤔 isn't the `mavenExtensions.hasExtension(MavenCoordinates.parseCoordinates(config.getMavenExtensionCustomCoordinates())` redundant ? ie either it's false in the previous condition so it'll be false there, or it's true and then this block will never get executed.

alextu (commented):
We would need to try out the upgrade path, after upgrading the plugin:
- Gradle builds:
there will be an intermediate state where some nodes will have the old env vars and old init script, some nodes will have the new env vars and new init script. 
- Maven builds:
there will be an intermediate state where some nodes will have `MAVEN_OPTS` set with the "old" sys props while some other nodes will `MAVEN_OPTS` set with the "new" sys props.
In both cases, there should not be a mix up and injection should still work

src/main/java/hudson/plugins/gradle/injection/InitScriptTemplate.java
leonard84: Would make things easier. I leave the decision up to you.

configuration-maven-extension/build.gradle.kts
```diff
@@ -13,17 +15,21 @@ group = "com.gradle"
...
+        url = URI.create("https://repo.grdev.net/artifactory/public")
+    }
 }
 
 val mvnExtension: Configuration by configurations.creating { isCanBeConsumed = true; isCanBeResolved = false }
 
 dependencies {
     compileOnly("org.apache.maven:maven-core:3.8.7")
-    compileOnly("com.gradle:gradle-enterprise-maven-extension:${gradleEnterpriseMavenExtensionVersion}")
+    compileOnly("com.gradle:develocity-maven-extension:${develocityMavenExtensionVersion}")
```
pshevche, line 32: ❗ We need a copy of https://github.com/jenkinsci/gradle-plugin/blob/master/configuration-maven-extension/src/main/java/com/gradle/jenkins/maven/extension/internal/DevelocityConfigurerListener.java#L14 which will implement `DevelocityListener`. Like in CCUD: [GE listener](https://github.com/gradle/common-custom-user-data-maven-extension/blob/main/src/main/java/com/gradle/CommonCustomUserDataGradleEnterpriseListener.java) and [DV listener](https://github.com/gradle/common-custom-user-data-maven-extension/blob/main/src/main/java/com/gradle/CommonCustomUserDataDevelocityListener.java). 

I think it is better to wait with this change until https://github.com/gradle/develocity-agent-adapters is released.

src/main/java/hudson/plugins/gradle/injection/DevelocityExceptionLogProcessor.java
```diff
@@ -16,6 +16,14 @@ public final class DevelocityExceptionLogProcessor extends AbstractGradleLogProc
...
         {
+            new DevelocityExceptionDetector.ByPrefix(
+                BuildToolType.GRADLE,
+                "Internal error in Develocity Gradle plugin:"
+            ),
+            new DevelocityExceptionDetector.ByPrefix(
+                BuildToolType.MAVEN,
+                "[ERROR] Internal error in Develocity Maven extension:"
+            ),
             new DevelocityExceptionDetector.ByPrefix(
```
pshevche, line 27: ❔ Why do we need this one? I think it is fine to always refer to the product as "Develocity".

src/main/java/hudson/plugins/gradle/injection/InitScriptTemplate.java
pshevche: Now I also feel like it would be better to just set the correct environment variables. In some plugins (e.g. in GitLab templates), it is a bit tricky to implement this replacement, so maybe script variables should just become our API for using the script.

pshevche (changes requested):

welandaz (approved):
Nice having this in! 🎉  
Left some comments.

welandaz (commented):
@leonard84 Please, also check failing tests

src/main/java/hudson/plugins/gradle/injection/DevelocityExceptionLogProcessor.java
leonard84: Old plugin versions will log the old string

src/main/java/hudson/plugins/gradle/injection/GradleBuildScanInjection.java
```diff
@@ -86,59 +75,59 @@ private static String getInitScriptDirectory(EnvVars envGlobal, EnvVars envCompu
...
 
-    private void injectInitScript(VirtualChannel channel, String initScriptDirectory) throws IOException, InterruptedException {
-        FilePath gradleInitScriptFile = getInitScriptFile(channel, initScriptDirectory);
+    private void injectInitScript(Node node, String initScriptDirectory) throws IOException, InterruptedException {
+        FilePath gradleInitScriptFile = getInitScriptFile(node.getChannel(), initScriptDirectory);
         if (initScriptChanged(gradleInitScriptFile)) {
             LOGGER.info("Injecting Gradle init script " + gradleInitScriptFile);
-
-            copyResourceToNode(gradleInitScriptFile, RESOURCE_INIT_SCRIPT_GRADLE);
+            writeToNode(gradleInitScriptFile, initScript.get());
```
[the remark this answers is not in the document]
leonard84: Because the script is templated and doesn't exist in the final form on disk or in a jar. Related to this comment https://github.com/jenkinsci/gradle-plugin/pull/409#discussion_r1528328438

src/main/java/hudson/plugins/gradle/injection/InitScriptTemplate.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.regex.Pattern;
+
+public class InitScriptTemplate {
+
+    private static final Map<String, String> REPLACEMENTS;
+
+    static {
+        Map<String, String> replacements = new HashMap<>();
+        Arrays.stream(InitScriptVariables.values())
+            .filter(v -> !v.getTemplateName().isEmpty())
```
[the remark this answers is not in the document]
leonard84: left over from a time where they could be empty, I will fix it

src/main/java/hudson/plugins/gradle/injection/InitScriptTemplate.java
```diff
@@ -0,0 +1,41 @@
...
+            .forEach(v -> replacements.put(v.getTemplateName(), v.getTemplateTargetName()));
+        REPLACEMENTS = Collections.unmodifiableMap(replacements);
+    }
+
+    public static String replacePlaceholders(String template, Map<String, String> replacements) {
+        Pattern pattern = Pattern.compile("getInputParam\\([\"']([^\"']+)[\"']\\)");
+        Matcher matcher = pattern.matcher(template);
+        StringBuffer buffer = new StringBuffer();
```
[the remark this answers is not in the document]
leonard84: Because `Matcher.appendReplacement` only accepts a `StringBuffer`, unfortunately.

src/main/resources/hudson/plugins/gradle/injection/init-script.gradle
```diff
@@ -82,70 +107,131 @@ def getInputParam = { String name ->
...
+                    if (isAtLeast(develocityPluginVersion, '2.1') && atLeastGradle5) {
+                        if (isAtLeast(develocityPluginVersion, '3.17')) {
+                            buildScanExtension.capture.fileFingerprints = develocityCaptureFileFingerprints
+                        } else if (isAtLeast(develocityPluginVersion, '3.7')) {
+                            buildScanExtension.capture.taskInputFiles = develocityCaptureFileFingerprints
+                        } else {
+                            buildScanExtension.captureTaskInputFiles = develocityCaptureFileFingerprints
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
leonard84: Ideally, you'd update the github-access version so I can use the updated script from there. The whole idea is to have a single version of this script. Ideally, in a dedicated repository, but that is not for the testing team to decide.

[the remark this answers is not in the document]
leonard84: According to @pshevche the change is already in the github-actions script.

src/main/java/hudson/plugins/gradle/injection/InitScriptTemplate.java
alextu: @welandaz what do you think?

welandaz: I'd also go with removing that prefix. I can't remember why it was introduced in the first place, but it seems unnecessary.

leonard84:
### Failing `BuildScanInjectionMavenCrossVersionTest` is due to 
> Develocity Maven extension version 1.21-rc-3 is a release candidate version. Release candidate versions are not accepted by this Develocity server.


### Failing `hudson.plugins.gradle.injection.MavenOptsDevelocityFilterTest`

@alextu / @welandaz I'm not sure how to properly implement the changes for `hudson.plugins.gradle.injection.MavenOptsDevelocityFilter`. Do we want to introduce multiple sets of `required` `SystemProperty.Key`? Currently, I've added both old and new versions, but that always means one set of `required` properties is missing when applied, resulting in all properties being removed. I'm not familiar enough with the implications to decide which fix would be appropriate here.

alextu:
Closing as it's been replaced by https://github.com/jenkinsci/gradle-plugin/pull/412

PR #411, opened 2024-03-19
Capture goal/task input files by default

welandaz:
Capturing of goal/task input files is now enabled by default. This is done by pre-selecting the corresponding checkbox in UI with the default value `true`.

The functionality is also only applied for cases where gradle plugin/maven extension isn't already applied.

### DIFF
--- a/src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
+++ b/src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
@@ -162,6 +162,9 @@ class BuildScanInjectionMavenIntegrationTest extends BaseMavenIntegrationTest {
             with(it.next()) {
                 it == '-Dgradle.enterprise.url=https://scans.gradle.com'
             }
+            with(it.next()) {
+                it == '-Dgradle.scan.captureGoalInputFiles=false'
+            }
             !it.hasNext()
         }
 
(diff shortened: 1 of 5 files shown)


src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
```diff
@@ -162,6 +162,9 @@ class BuildScanInjectionMavenIntegrationTest extends BaseMavenIntegrationTest {
             with(it.next()) {
                 it == '-Dgradle.enterprise.url=https://scans.gradle.com'
             }
+            with(it.next()) {
+                it == '-Dgradle.scan.captureGoalInputFiles=false'
```
alextu, line 166: 🤔 Shouldn't it be `gradle.scan.captureGoalInputFiles=true` by default ?

alextu (approved):

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
welandaz: In this specific test, the `turnOnBuildInjectionAndRestart` method sets it to `false` by default.
I've changed the default value to `true`

alextu:
Closing this one as I've merged those changes into https://github.com/jenkinsci/gradle-plugin/pull/412

PR #418, opened 2024-04-05
Print only necessary variables for assertions

welandaz:
This PR removes the output of whole environment during test jobs and prints only necessary variables for assertions

### DIFF
--- a/src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
+++ b/src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
@@ -676,10 +676,8 @@ node {
             $replacedMavenSetup
                 git branch: 'main', url: 'https://github.com/c00ler/simple-maven-project'
                 if (isUnix()) {
-                    sh "echo \$MAVEN_OPTS"
                     sh "mvn package -B"
                 } else {
-                    bat "echo %MAVEN_OPTS%"
                     bat "mvn package -B"
                 }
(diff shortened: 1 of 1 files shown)


src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
```diff
@@ -293,14 +293,32 @@ class BuildScanInjectionMavenIntegrationTest extends BaseMavenIntegrationTest {
...
+}
+""", false))
 
         when:
         def build = j.buildAndAssertSuccess(pipelineJob)
 
         then:
-        j.assertLogContains("GRADLE_ENTERPRISE_ACCESS_KEY=scans.gradle.com=secret", build)
-        j.assertLogContains("DEVELOCITY_ACCESS_KEY=scans.gradle.com=secret", build)
+        j.assertLogContains("scans.gradle.com=secret", build)
```
alextu, line 321: I would keep the assertions and instead echo `GRADLE_ENTERPRISE_ACCESS_KEY=$GRADLE_ENTERPRISE_ACCESS_KEY` in the build, same for the other env vars

alextu (approved):

src/test/groovy/hudson/plugins/gradle/injection/BuildScanInjectionMavenIntegrationTest.groovy
welandaz: Agree, changed

PR #426, opened 2024-04-11
Use short-lived tokens instead of Access key

alextu:
Generate a short lived token before each build and use it in the env var instead of the access key.
A new UI parameter is introduced to control the tokens expiry:

image

### Testing done

```[tasklist]
### Submitter checklist
```


welandaz (approved):
Nice! 🎉

src/main/java/hudson/plugins/gradle/injection/BuildScanEnvironmentContributor.java
```diff
@@ -23,15 +25,38 @@
...
         Secret secretKey = InjectionConfig.get().getAccessKey();
         Secret secretPassword = InjectionConfig.get().getGradlePluginRepositoryPassword();
         if ((secretKey == null && secretPassword == null) || alreadyExecuted(run)) {
             return;
         }
 
-        run.addAction(DevelocityParametersAction.of(new DevelocityLogger(listener), secretKey, secretPassword));
+        Secret shortLivedToken = null;
+        DevelocityLogger logger = new DevelocityLogger(listener);
+        if (secretKey != null && DevelocityAccessKey.isValid(secretKey.getPlainText())) {
```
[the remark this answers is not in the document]
alextu: Yeah, this double-check was already present, I just moved it there to make the test pass

src/main/java/hudson/plugins/gradle/injection/BuildScanEnvironmentContributor.java
```diff
@@ -23,15 +25,38 @@
...
+            shortLivedToken = tokenClient.get(InjectionConfig.get().getServer(),
+                    DevelocityAccessKey.parse(secretKey.getPlainText()),
```
bigdaz, line 51: Unless I'm mistaken, this `parse` method assumes that the access key contains a single `server=key` entry, but the `isValid` method above will pass for a key with `server1=key1;server2=key2`

alextu: You're right, then we need to extract the hostname from the DV server URL and try to find the corresponding key

src/main/java/hudson/plugins/gradle/injection/BuildScanEnvironmentContributor.java
```diff
@@ -9,29 +10,78 @@
...
+            logger.error("Develocity access key format is not valid");
+            return null;
+        }
+        Optional<DevelocityAccessKey> accessKey = DevelocityAccessKey.parse(secretKey.getPlainText(), hostname);
+        return accessKey
+            .map(k ->
+                tokenClient.get(serverUrl, k, InjectionConfig.get().getShortLivedTokenExpiry()))
+            .filter(Optional::isPresent)
+            .map(k -> Secret.fromString(k.get().getRawAccessKey()))
+            .orElse(null);
```
jthurne, line 76: If we fail to get a short-lived access token, do we want to fail the build, or fall back to some reasonable behavior?

In our own dogfooding case, we decided to not set the access key and let the build fail to use various Develocity features like build caching and build scan publishing. However, we had to explicitly disable Test Distribution (which will otherwise fail if it fails to connect to Develocity). I wonder if we should do something similar here.

I also wonder if it would make sense to add a configuration option to allow users to decide if they want the build to fail if it can't get an access token, or to continue without Develocity.

src/main/java/hudson/plugins/gradle/injection/token/ShortLivedTokenClient.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @SuppressFBWarnings("NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE")
+    public Optional<DevelocityAccessKey> get(String server, DevelocityAccessKey accessKey, @Nullable Integer expiry) {
+        String url = normalize(server) + "api/auth/token";
+        if (expiry != null) {
+            url = url + "?expiresInHours=" + expiry;
+        }
+
+        Request request = new Request.Builder()
```
jthurne, line 39: I recommend implementing some form of a retry mechanism. That way, if the Develocity instance is down briefly, the build has a chance of still working.

We found a retry mechanism to be quite helpful in our own dogfooding of short-lived access tokens.

welandaz (approved):

basil:
Nice fix! [JENKINS-73082](https://issues.jenkins.io/browse/JENKINS-73082) would also be nice to fix.

PR #454, opened 2024-06-09
Use Library Plugins for okhttp, jackson

sghill:
Hi! This updates the plugin to depend on the api-plugin versions of okhttp and jackson. This is a [recommended approach](https://www.jenkins.io/doc/developer/plugin-development/dependencies-and-class-loading/#using-library-wrapper-plugins) to code sharing.

I have a few plugins that depend on `gradle` and `okhttp-api`. I'm currently pinning back to Gradle 2.11 because 2.12 brings in its own copy of okhttp (https://github.com/jenkinsci/gradle-plugin/pull/426), causing runtime errors on startup.

While `okhttp-api` alone would fix the problem for me, I included jackson2 because it has the same guidance.

Thanks for taking a look!

### Testing done

Existing tests pass after the change.

### Submitter checklist

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -92,6 +92,7 @@ val gradlePluginJpi: Configuration by configurations.creating { isCanBeConsumed 
 dependencies {
     api(platform("io.jenkins.tools.bom:bom-${coreBaseVersion}.x:${coreBomVersion}"))
 
+    implementation("org.jenkins-ci.plugins:jackson2-api")
     implementation("org.jenkins-ci.plugins:structs")
     implementation("org.jenkins-ci.plugins.workflow:workflow-api")
     implementation("org.jenkins-ci.plugins.workflow:workflow-cps")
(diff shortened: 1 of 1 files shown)


sghill:
@alextu (author of original PR) is this something you'd be able to take a look at?

alextu:
Thanks, for your contribution, will take a look today

alextu:
When I build the plugin on your branch, it depends on okhttp 4.9.2 and Jackson 2.13.3 (which has vulnerabilities).
So to get newer versions we need to depend on the latest Jenkins bom, right? 
This would bump the min Jenkins version requirement, which we don't want to do every week or so when a new version of Jackson or Okhttp is out. Even If I do bump to latest `2.426.x` bom, I get a build error I would need to investigate (`Could not resolve all task dependencies for configuration ':annotationProcessor'`)

As an alternative, If I fix wrapper version plugins like to `org.jenkins-ci.plugins:jackson2-api:2.17.0-379.v02de8ec9f64c` and `io.jenkins.plugins:okhttp-api:4.11.0-172.vda_da_1feeb_c6e`, I get an obscure `java.lang.UnsupportedOperationException: This feature requires ASM7` when running `animalSniffer` task 🤷‍♂️

sghill:
True, but getting outdated versions is part of keeping the bom requirement lower.

The version declared by the plugin is only setting the minimum, not necessarily what users will run with. For example I'm running with the latest jackson2-api and okhttp-api, so that's what I'm going to get at runtime.

alextu:
Ah I see, the dependencies become just "runtime" dependencies and dependency management is handled by the Jenkins runtime. @welandaz, ok on your side?

welandaz:
> Ah I see, the dependencies become just "runtime" dependencies and dependency management is handled by the Jenkins runtime. @welandaz, ok on your side?

Yes, fine by me

welandaz (approved):

sghill:
Thank you both! Looking forward to upgrading when this is out.

PR #476, opened 2024-08-30
Download Maven extensions instead of bundling them

welandaz:
This PR removes the bundling of Develocity and CCUD Maven extensions.
Instead, the extensions are now downloaded from Maven Central. This enables the user to specify the version of the extensions they need and keep using the same plugin version.

This change is not backwards compatible and users would need to explicitly specify the extension versions after an upgrade.

Screenshot 2024-08-30 at 09 21 13


guylabs (approved):

README.adoc
```diff
@@ -292,36 +292,11 @@ The compatibility of the specified version with Develocity can be found https://
...
-|2.0                            | 1.15.4                                             | 1.11.1                                                    | 2022.3
-|1.40                           | 1.15.3                                             | 1.11.1                                                    | 2022.3
-|< 1.40                         | *Not supported*                                    | *Not supported*                                           | *Not supported*
-|===
+For Maven builds the version used for the Develocity Maven extension is defined in the `Develocity Maven extension version` field in the `Maven settings` section of the configuration form.
+See <<Enable auto-injection>> for details.
+The compatibility of the specified version with Develocity can be found https://docs.gradle.com/enterprise/compatibility/#maven_extension[here].
+
+For the optional Common Custom User Data Maven extension which is defined the same form, you can see the compatibility of the specified version with the Develocity Maven extension https://github.com/gradle/common-custom-user-data-maven-extension#version-compatibility[here].
```
alextu, line 300: Shouldn't we put a warning note explaining that the injection will not work after upgrading referencing https://github.com/jenkins-infra/helpdesk/issues/4263.

alextu (approved):
Have you been able to manually test the upgrade path when the "old" plugin has the injection enabled?

welandaz:
> Have you been able to manually test the upgrade path when the "old" plugin has the injection enabled?

Yes, the injection get's disabled after an upgrade as the configuration fields are no longer there and it triggers the "disable injection" code path. User's would need to specify the extension versions explicitly and re-enable it again

README.adoc
welandaz: I thought about adding it to the release notes only.
Do you think we should mention it in the README as well?

alextu: Yes, seems like a breaking change that should stay, see for example what we did when requiring Develocity 2024.1 https://github.com/gradle/develocity-gitlab-templates?tab=readme-ov-file#requirements

welandaz: Agree, although I'd update a Readme at the time of release, as it might be confusing to have this mentioned already without a version available

PR #488, opened 2024-09-16
Download extensions only once and if the version changes

welandaz:
This change will download Maven extensions only once to the controller node and then use the same copying mechanism we originally had. The only difference is that before we had a predefined location where the extensions where stored (as they were bundled), and now they will be re-downloaded if needed into `jenkins-gradle-plugin/cache` on the  controller node.

The download happens for the first time when injection get's enabled or when a user updates the version they want to inject. We now store a metadata file with a downloaded version and a digest of the JAR which we use for a comparison to determine if we need to copy it to an agent node.


src/main/java/hudson/plugins/gradle/injection/MavenExtensionDownloadHandler.java
```diff
@@ -0,0 +1,97 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class MavenExtensionDownloadHandler implements MavenInjectionAware {
```
welandaz, line 19: I tested this logic manually, but I'm going to add a test as part of this PR for this specific class

welandaz: Test added

src/main/java/hudson/plugins/gradle/injection/DevelocityComputerListener.java
```diff
@@ -21,29 +24,51 @@ public class DevelocityComputerListener extends ComputerListener {
...
         try {
+            InjectionConfig injectionConfig = injectionConfigSupplier.get();
             EnvVars globalEnvVars = computer.buildEnvironment(listener);
-            if (InjectionUtil.globalAutoInjectionCheckEnabled(globalEnvVars) && isFeatureDisabled()) {
+            if (InjectionUtil.globalAutoInjectionCheckEnabled(globalEnvVars) && injectionConfig.isDisabled()) {
                 return;
             }
 
-            injector.inject(computer, globalEnvVars);
+            Map<MavenExtension, String> extensionsDigest = mavenExtensionDownloadHandler.ensureExtensionsDownloaded(
```
alextu, line 63: This can be called concurrently (multiple agents launching at the same time). If somehow the extension was not downloaded after saving the configuration, it could get corrupted when downloaded multiple times. If we think it's a realistic issue we can just have `ensureExtensionDownloaded` making sure the extension is there without downloading it (so no injection in that case). After all now it has to be an explicit action to enable the Maven injection. Wdyt?

src/main/java/hudson/plugins/gradle/injection/MavenExtensionDownloadHandler.java
```diff
@@ -0,0 +1,103 @@
...
+
+    public static final String DOWNLOAD_CACHE_DIR = "jenkins-gradle-plugin/cache";
+
+    public Map<MavenExtension, String> ensureExtensionsDownloaded(File root, InjectionConfig injectionConfig) throws IOException {
+        if (!isInjectionDisabledGlobally(injectionConfig)) {
+            Map<MavenExtension, String> extensionsDigest = new HashMap<>();
+            Path parent = root.toPath().resolve(DOWNLOAD_CACHE_DIR);
```
alextu, line 28:
```suggestion
            Path cacheDir = root.toPath().resolve(DOWNLOAD_CACHE_DIR);
```

src/main/java/hudson/plugins/gradle/injection/MavenExtensionDownloadHandler.java
```diff
@@ -0,0 +1,103 @@
...
+                extensionsDigest.put(MavenExtension.CCUD, getExtensionDigest(injectionConfig, parent, MavenExtension.CCUD));
+            }
+
+            return extensionsDigest;
+        }
+
+        return Collections.emptyMap();
+    }
+
+    private static String getExtensionDigest(InjectionConfig injectionConfig, Path parent, MavenExtension extension) throws IOException {
```
alextu, line 43:
```suggestion
    private static String getExtensionDigest(InjectionConfig injectionConfig, Path cacheDir, MavenExtension extension) throws IOException {
```

src/main/java/hudson/plugins/gradle/injection/MavenExtensionDownloadHandler.java
```diff
@@ -0,0 +1,103 @@
...
+                return extensionDigest;
+            }
+        } else {
+            return downloadExtension(injectionConfig, parent, extension, metadataFile);
+        }
+    }
+
+    private static String downloadExtension(
+            InjectionConfig injectionConfig,
+            Path parent,
```
alextu, line 61:
```suggestion
            Path cacheDir,
```

src/test/groovy/hudson/plugins/gradle/injection/MavenExtensionDownloadHandlerTest.groovy
```diff
@@ -0,0 +1,90 @@
...
+
+class MavenExtensionDownloadHandlerTest extends Specification {
+
+    @Rule
+    public TemporaryFolder tempFolder = new TemporaryFolder()
+
+    @Subject
+    private final MavenExtensionDownloadHandler mavenExtensionDownloadHandler = new MavenExtensionDownloadHandler()
+
+    def 'extensions are not redonwloaded if config has not changed'() {
```
alextu, line 16:
```suggestion
    def 'extensions are not redownloaded if config has not changed'() {
```

src/test/groovy/hudson/plugins/gradle/injection/MavenExtensionDownloadHandlerTest.groovy
```diff
@@ -0,0 +1,90 @@
...
+        def originalCcudDigest = extensionsDownloaded.get(MavenExtension.CCUD)
+
+        originalDevelocityDigest != null && originalCcudDigest != null
+
+        when:
+        def sameExtensions = mavenExtensionDownloadHandler.ensureExtensionsDownloaded(controllerFolder, originalConfig)
+
+        then:
+        sameExtensions.size() == 2
+        sameExtensions.get(MavenExtension.DEVELOCITY) == originalDevelocityDigest && sameExtensions.get(MavenExtension.CCUD) == originalCcudDigest
```
alextu, line 43: 🤔 how does it test if it's not re-downloaded?

alextu (approved):

src/test/groovy/hudson/plugins/gradle/injection/MavenExtensionDownloadHandlerTest.groovy
welandaz: Just by verifying that the digest of the one downloaded before hasn't changed.
Or do you mean that it can still download it, just the same version again, which will produce same digest?

alextu: yes, exactly

src/main/java/hudson/plugins/gradle/injection/DevelocityComputerListener.java
welandaz: I've removed downloading on agent's onOnline event as we discussed.

Everything else stays as it's sequential and synchronized by Jenkins

src/main/java/hudson/plugins/gradle/injection/MavenExtensionDownloadHandler.java
welandaz: Fixed

src/main/java/hudson/plugins/gradle/injection/MavenExtensionDownloadHandler.java
welandaz: Fixed

src/main/java/hudson/plugins/gradle/injection/MavenExtensionDownloadHandler.java
welandaz: Fixed

src/test/groovy/hudson/plugins/gradle/injection/MavenExtensionDownloadHandlerTest.groovy
welandaz: Added additional check for lastModified

alextu (approved):

PR #495, opened 2024-10-08
Fix failure with configuration-as-code plugin when upgrading

alextu:
https://issues.jenkins.io/browse/JENKINS-73843

`injectMavenExtension` and `injectCcudExtension` have been removed in 2.13. It causes a warning in Jenkins admin UI after upgrading, it causes a hard failure if using the Jenkins configuration-as-code plugin with a yaml configuration referencing those removed attributes.

This PR just keeps those attributes as transient, till the end of times I guess.

### Testing done
- Integration test
- Manually tested the upgrade with a pre-configured JCasC YAML referencing `injectMavenExtension` and `injectCcudExtension`.

### Submitter checklist

### DIFF
--- a/src/main/java/hudson/plugins/gradle/injection/InjectionConfig.java
+++ b/src/main/java/hudson/plugins/gradle/injection/InjectionConfig.java
@@ -102,6 +102,9 @@ public class InjectionConfig extends GlobalConfiguration {
     // Legacy property that is not used anymore
     private transient String injectionVcsRepositoryPatterns;
     private VcsRepositoryFilter parsedVcsRepositoryFilter = VcsRepositoryFilter.EMPTY;
+    // Legacy properties, kept for reading old configurations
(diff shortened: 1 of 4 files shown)


src/main/java/hudson/plugins/gradle/injection/InjectionConfig.java
```diff
@@ -102,6 +102,9 @@ public class InjectionConfig extends GlobalConfiguration {
     // Legacy property that is not used anymore
     private transient String injectionVcsRepositoryPatterns;
     private VcsRepositoryFilter parsedVcsRepositoryFilter = VcsRepositoryFilter.EMPTY;
+    // Legacy properties, kept for reading old configurations
```
welandaz, line 105: I think we should mention in the comment that this is only for JCasC plugin

welandaz (approved):

PR #530, opened 2025-02-20
Require 2.479.x or newer and migrate from EE 8 to EE 9

basil:
Migrate from deprecated EE 8 functionality to non-deprecated EE 9 equivalents. See https://github.com/jenkinsci/lockable-resources-plugin/pull/743#issuecomment-2598199297 for more background.

The actual migration from EE 8 to EE 9 is small, but the requirement of 2.479.x or newer (which requires Java 17 or newer) is nontrivial when it comes to updating this plugin's tests. I have made a minimal effort to update the build, but I am leaving the updating of the tests to the plugin maintainers.

### Testing done

I did some light testing locally but did not get all the automated tests to pass. (For that matter, https://ci.jenkins.io/job/Plugins/job/gradle-plugin/job/master/ is also failing at this moment.) I am leaving the updating of the tests to the plugin maintainers, who are more familiar with this plugin and Gradle Java requirements.

### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPluginWithGradle(jdkVersions: ['11'], timeout: 180)
+buildPluginWithGradle(jdkVersions: ['17'], timeout: 180)
--- a/acceptance-tests/build.gradle.kts
+++ b/acceptance-tests/build.gradle.kts
@@ -14,7 +14,7 @@ java {
     // Only used for compilation. We don't rely on toolchain for running the tests,
     // as Jenkins ATH doesn't allow to specify the JAVA_HOME.
     toolchain {
-        languageVersion.set(JavaLanguageVersion.of(11))
+        languageVersion.set(JavaLanguageVersion.of(17))
     }
 }
 
@@ -32,9 +32,9 @@ val gradlePlugin: Configuration by configurations.creating { isCanBeConsumed = f
 
 dependencies {
     // same version as used by ATH
-    annotationProcessor("org.jenkins-ci:annotation-indexer:1.12")
+    annotationProcessor("org.jenkins-ci:annotation-indexer:1.18")
 
-    implementation("org.jenkins-ci:acceptance-test-harness:5740.vd30f30408987")
+    implementation("org.jenkins-ci:acceptance-test-harness:6133.v358d9a_47674f")
 
     testImplementation(platform("io.netty:netty-bom:4.1.118.Final"))
     testImplementation("io.ratpack:ratpack-test:2.0.0-rc-1")
@@ -45,7 +45,7 @@ dependencies {
 val jenkinsVersions = listOf(
     JenkinsVersion.LATEST,
     JenkinsVersion.LATEST_LTS,
-    JenkinsVersion.V2_440
+    JenkinsVersion.V2_479
 )
 
 jenkinsVersions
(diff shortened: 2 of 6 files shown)


welandaz:
hi @basil !
Thanks a lot for the PR!

We've been careful with Jenkins version updates as we encountered cases before where users couldn't use newer version of the plugin as it required a Jenkins update (this is particularly true for Develocity customers who use auto-injection). 

The result of this PR would be a bump of JDK 11 to JDK 17 which we're a bit hesitant to do now. 

Does Jenkins community now have a requirement that all or most plugins push for JDK 17 adoption (or EE 9 migration)? If not, we'd delay this until absolutely necessary.

basil:
@welandez https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ describes our general guidance about Jenkins core baselines. We are currently recommending 2.462.3 and 2.479.3 as core baselines for the majority of plugins. Our guidance for the most conservative plugins that want to support the oldest core baseline they can is to choose the oldest release still supported by the Update Center. The Update Center supports releases for just about a year, as mentioned on that page:

> Do not use versions no longer supported by the update center, which is currently anything older than 2.440 for weekly releases, and 2.426.2 for LTS releases.

The EE 9 update shipped in September 2024, so we'd want even the most conservative plugins to be on EE 9 by around September 2025. While there is no urgent rush, the compatibility layer for EE 8 was a significant source of technical debt, and we want to get the ecosystem onto EE 9 so that we can reduce the surface area of the compatibility layer in future Jenkins core releases.

basil:
@welandaz https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/ describes our general guidance about Jenkins core baselines. We are currently recommending 2.462.3 and 2.479.3 as core baselines for the majority of plugins. Our guidance for the most conservative plugins that want to support the oldest core baseline they can is to choose the oldest release still supported by the Update Center. The Update Center supports releases for just about a year, as mentioned on that page:

> Do not use versions no longer supported by the update center, which is currently anything older than 2.440 for weekly releases, and 2.426.2 for LTS releases.

The EE 9 update shipped in September 2024, so we'd want even the most conservative plugins to be on EE 9 by around September 2025. While there is no urgent rush, the compatibility layer for EE 8 was a significant source of technical debt, and we want to get the ecosystem onto EE 9 so that we can reduce the surface area of the compatibility layer in future Jenkins core releases.

welandaz:
@basil Thanks a lot for the context!

Would you be fine then with us postponing this update and do it closer to September 2025? 
I'd still keep the PR open, but we would proceed with it later once we know what's up ahead for the plugin in the coming months.

basil:
@welandaz That seems fine. Are the Gradle folks OK with updating this plugin's test suite for Java 17? When I took a quick look at the failures, they didn't seem Jenkins-specific. Rather, it looked like some tests were running older versions of Gradle/Develocity and those started failing on Java 17. Since I'm not familiar with your product lifecycle and testing requirements, it seems best to leave this to the plugin maintainers.

welandaz:
@basil You mean just to have the tests with JDK 17, but leave everything else as is? I think we can do that. 
Develocity plugins/extensions are JDK 8 compatible and are tested with latest JDK as well, so they shouldn't be the issue.
We'd need to check tests with older Gradle versions as those wouldn't work with JDK 17; this shouldn't be a blocked for the update.

basil:
@welandaz Thanks! That sounds great, so I am closing this PR, and the Gradle folks can pick up this effort at some point in the future whenever they desire to require 2.479.x or newer.

PR #559, opened 2025-05-16
Rename input files to file fingerprints in UI

welandaz:
A simple rename of input files to file fingerprints in UI as it reflects the rename done on build agents' side a while back.


alextu (approved):
Can we rename the config parameter without too much hassle?

welandaz:
> Can we rename the config parameter without too much hassle?

If you're talking about the fields themselves in the `InjectionConfig`, then that would mean migration :(

alextu:
Yeah, let's not do this

